PR #50, opened 2020-05-17
Store instance in ValidationError

macisamuele:
The goal of this PR is to unify the content of all the possible `ValidationError` instances.
In order to do so:
 * instance is always stored into `ValidationError` and not in `ValidationErrorKind`
 * `ValidationErrorKind` contains only the details of the specific kind of error (and all are structs instead of mixture of struct/tuples)
 * `ValidationError` stores a `CoW` instance of the instance in order to reduce the amount of `Clone::clone` invocations

The idea, in the long run, would be to ensure that `ValidationError` stores only references to the instance that caused the validation error by always avoiding cloning them (this will have memory and runtime benefits).
Not doing it in this PR due to PR size (`ValidationError::into_owned` should ideally be removed)

### DIFF
--- a/src/compilation.rs
+++ b/src/compilation.rs
@@ -1,10 +1,13 @@
 //! Schema compilation.
 //! The main idea is to compile the input JSON Schema to a validators tree that will contain
 //! everything needed to perform such validation in runtime.
-use crate::error::{CompilationError, ErrorIterator};
-use crate::keywords::Validators;
-use crate::resolver::Resolver;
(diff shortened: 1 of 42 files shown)


Stranger6667:
> The idea, in the long run, would be to ensure that ValidationError stores only references to the instance that caused the validation error by always avoiding cloning them (this will have memory and runtime benefits).

This is amazing! I tried to do so before but failed :) Glad to see these changes coming

Issue #30, opened 2020-03-29
Python interface


Stranger6667:
PyO3

As proposed by @macisamuele , this library might help - https://github.com/macisamuele/json-trait-rs

CFFI implementation that I started is far from being done, there are some weird edge cases that I don't know how to handle. E.g. calling `PyLong_FromLongLong` on a big integer produces more corruption :/

macisamuele:
@Stranger6667 I'll try to update all the keywords in order to allow validation of a generic JsonType. I'll try to compare it with the benchmarks to identify possible regressions or additional issues

Stranger6667:
Do you mean non-serde types?

macisamuele:
Ideally (at least this what I was experimenting with) the `Validate` trait should be modified into something like
```rust
trait Validate<J: JsonType> {
    fn validate(&self, schema: JsonSchema<J>, value: &J) -> ...;
    fn is_valid(&self, schema: JsonSchema<J>, value: &J) -> bool;
}
```
By doing so we might validate for any type `J` that implements `JsonType` trait.

If by doing it we don't introduce performance regressions then we might simplify the integration with foreign platforms (which might be Python or any other FFI) that adhere to `JsonType` trait.

Still this is a very preliminary idea, as I need to finalise the test branch to see the feasibility (especially comparing performances).

Once I have results (maybe over the weekend) I'll post something in here.

Stranger6667:
Done in #54.

Sdist & wheels for manylinux2010 are [here](https://pypi.org/project/jsonschema-rs/#files)

PR #51, opened 2020-05-17
Use serde_json APIs instead relying on value enum

macisamuele:
The goal of this PR is to decouple (as much as possible) from `serde_json`.
This is mostly needed to simplify the integration of `json-trait-rs` (#30).

I'm publishing this for initial review as it will be the base of the work for testing `json-trait-rs`.

I'm very sorry for the size of the PR ... :(

### DIFF
--- a/src/keywords/helpers.rs
+++ b/src/keywords/helpers.rs
@@ -1,8 +1,19 @@
 use serde_json::Value;
 
+fn are_f64_equal(left: f64, right: f64) -> bool {
+    (left - right).abs() < 3. * f64::EPSILON
+}
+
 pub(crate) fn equal(left: &Value, right: &Value) -> bool {
-    match (left, right) {
-        (Value::Number(left), Value::Number(right)) => left.as_f64() == right.as_f64(),
-        (_, _) => left == right,
+    if left == right {
+        true
+    } else if let Some(left_num) = left.as_f64() {
+        if let Some(right_num) = right.as_f64() {
+            are_f64_equal(left_num, right_num)
+        } else {
+            false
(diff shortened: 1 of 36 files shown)


src/keywords/helpers.rs
```diff
@@ -1,8 +1,19 @@
 use serde_json::Value;
 
+fn are_f64_equal(left: f64, right: f64) -> bool {
+    (left - right).abs() < 3. * f64::EPSILON
```
Stranger6667, line 4: Why not comparison of `left_num` and `right_num`?

Stranger6667:
> I'm very sorry for the size of the PR ... :(

No probs :) 

I see that it does some negative impact on performance (however it is small in absolute numbers), e.g. with some `type` benches it has >+100% time change. I understand that it might be a cost to support non-serde case, but is there any way to avoid this? Or what ways do we have to improve it in the future?

macisamuele:
I was noticing the performance hit, but overall absolute number were very low.

Still I'm not 100% sure that json-trait-ts integration is feasible and reasonable. So I would eventually suggest to not merge this yet until some further tests are done

Stranger6667:
Yep, let's take a look at how it could work performance-wise in comparison with other approaches.
I opened drafts #53 and #54 for comparison (I'd be glad if you can take a look). Will try to clean them up so it could be measured better

Cheers

macisamuele:
@Stranger6667 I'm going to drop this for now as it does have a massive amount of merge conflicts as well as #86 might also simplify this effort as it would require a very localised diff and not on all the keywords.

PR #62, opened 2020-05-19
Improve compilation performance

Stranger6667:
1. Avoid string allocation during scope parsing & do not re-parse DEFAULT_ROOT_URL

Before, the string was cloned anyway - either DEFAULT_ROOT_URL or the one from the schema and then passed to `Url::parse` where it was cloned again.

Benches:
`boolean_false compile false`: -72% (212.48 ns -> 59.542 ns)
`jsonschema big schema compile`: -7.1% (3.4517 us -> 3.2085 us)
`jsonschema small schema compile`: -5.6% (3.3137 us -> 3.1238 us)

This behavior is reproducible over other benchmarks. It will have the most effect on the (re)compilation of small schemas + frequent use `jsonschema::is_valid` without reusing compiled schema.

2. `$ref` and boolean validators are compiled without calling `push` on a vector, i.e. the vector is created more efficiently

It gives ~9.6% speedup for boolean validators and really small speedup for `$ref` (probably need a better benchmark). There is a visible improvement in the `perf` output.

3. Avoid `to_string` call during format validators compilation

It gives ~11% improvement

4. Mark all `compile` calls with `#[inline]`

On validators that don't involve regex compilation, it gave from 3 to 25% improvement

5. Reusing `DEFAULT_SCOPE` in `resolve_fragment`

Not really about compilation, but it is 25-30% of improvement during reference resolving on various benches

6. Replacing `FormatValidator` with generated structs

Improves compilation for this keyword by 7-8% and makes validation slightly faster (probably due to removing a func call overhead). But for `validate` there is an extra `match`, which seems to be slightly faster than a function call overhead. I'll cleanup duplication later


Stranger6667:
So, for some more or less realistic schemas, this PR gives 8-12% improvement in the compilation time and from 40% to 80% improvement for small single-keyword schemas

PR #74, opened 2020-05-21
fix: Make `is_valid` return `Result`

### LINKED ISSUE #56: `is_valid` should return Result
Because it can fail on SchemaError

Stranger6667:
Resolves #56

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,18 +57,20 @@ extern crate lazy_static;
 
 /// Validates `instance` against `schema`. Draft version is detected automatically.
 /// ```rust
-/// use jsonschema::is_valid;
+/// use jsonschema::{is_valid, CompilationError};
 /// use serde_json::json;
 ///
-///
-/// let schema = json!({"maxLength": 5});
-/// let instance = json!("foo");
-/// assert!(is_valid(&schema, &instance));
+/// fn main() -> Result<(), CompilationError> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -57,18 +57,20 @@ extern crate lazy_static;
...
+///    let instance = json!("foo");
+///    assert!(is_valid(&schema, &instance)?);
+///    Ok(())
+/// }
 /// ```
 #[inline]
-pub fn is_valid(schema: &Value, instance: &Value) -> bool {
-    let compiled = JSONSchema::compile(schema, None).expect("Invalid schema");
-    compiled.is_valid(instance)
+pub fn is_valid(schema: &Value, instance: &Value) -> Result<bool, CompilationError> {
```
macisamuele, line 71: I'm unsure about this change.
From API ergonomics it seems weird to have an `is_*` not returning a boolean.

I would suggest to return false in case a `CompilationError` occurs (because the `instance` is not valid against and invalid `schema`) and eventually logging the details of incorrect schema

Stranger6667: Indeed I am not sure either, but I do think that we should be explicit that this operation has 3 possible outcomes and it may be more confusing to return a result indicating that the instance is not valid, however, the schema is invalid and not the instance. Maybe the name does not completely reflect its nature

PR #77, opened 2020-05-22
chore: Convert gitmodule to https

leoschwarz:
I just noticed this issue, while compiling a fork of your crate without TLS because I want to statically link it with musl. If you don't want to include this, that's fine from my side but it might make it easier for others to depend on git versions of your crate & forks.


Stranger6667:
Hi @leoschwarz !

Sure thing :) Thank you for your contribution!

Issue #63, opened 2020-05-20
Group validators by the input type


Stranger6667:
The idea is to store validators in groups by the input type, e.g. all validators that can be applied to a number, object, array, string, etc. 

What we can get from it

**Less pattern matching on the matching type**

Consider this schema: `{"minimum": 1, "maximum": 10}`

Essentially we have 2 validators that together roughly do the following:

```rust
if let Value::Number(item) = instance {
    let item = item.as_f64().unwrap();
    if item < self.limit {
        return false;
    }
}
if let Value::Number(item) = instance {
    let item = item.as_f64().unwrap();
    if item > self.limit {
        return false;
    }
}
```

Pattern matching twice, and `item.as_f64().unwrap()` twice. Instead, we can do on the root validation method (and in nodes where it is appropriate):

```rust
... // some common validators for any type here
match instance {
    Value::Number(item) => {
        let item = item.as_f64().unwrap();
        // first validator inlined for illustration
        if item < self.limit {
            return false;
        };
        if item > self.limit {
            return false;
        }
        true
    }
    ...
}
```

In this arm, we can apply `exclusiveMaximum`, `exclusiveMinimum`, `minimum`, `maximum`, and `multipleOf`.

**Much simpler validators**

Instead of this:

```rust
    fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
        if let Value::Number(item) = instance {
            let item = item.as_f64().unwrap();
            if item < self.limit {
                return false;
            }
        }
        true
    }
```

we can do this:

```rust
    fn is_valid(&self, item: f64) -> bool {
        item < self.limit
    }
```

And there is no need to pass a not used reference to `JSONSchema` instance. The same simplification can be applied to the `validate` method.

**Faster execution for not-matching types**

Currently, if we pass `null` to the validator above, we'll still call both of them in a loop. and they both will return `true`. With that idea, there will be only 1 pattern matching in the root + maybe some small checks which I'll describe below

**More insights where to apply parallel execution**

We can know for sure that there is no point to apply any parallel execution for numeric validators, since they are fast and there are only 5 of them. In other words, the surface of possibilities will be more visible (only applicable to arrays and objects) and smaller.

As a downside, I see that there could be some extra logic to iterate over two vectors (common & specific validators) which may have higher overhead for some small schemas with a single keyword

Also, the implementation will require splitting to multiple traits.

But anyway, this option is worth exploring, maybe some other optimizations will be more visible on the way

[...]

PR #78, opened 2020-05-22
Use BitMaps to validate multiple types

macisamuele:
The goal of this PR is to improve memory utilisation of `MultipleTypesValidator` as well as improving validation performance.

The validation performance is achieved by:
 * replacing iteration over a `Vec` instance (data are into the heap) with bit operations
 * ensuring that errors are as small as possible

Performance improvements have been measured by running the following
```bash
rm -rf target/criterion
git co -f 4cf7d8d
cargo bench "type_multiple"
git co -f 9437e9d
sleep 60  # To allow cooling down the laptop
cargo bench "type_multiple"
```
The raw output is available [here](https://gist.github.com/macisamuele/c01319c524b831061be0285e18458503)
Summary:
 * ~21% improvement in compile time
 * at least 12% improvement in validation time (it goes up to 34% as well)

NOTE: I've run a subset of benchmarks only selecting the relevant to multiple types

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use crate::primitive_type::PrimitiveType;
 use serde_json::Value;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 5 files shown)


Stranger6667:
Thank you Samuele!
It is an ingenious trick and I like it a lot :) Great!

What do you think if we can apply it somewhere else as well?

macisamuele:
I've noticed this "easy win" while working on [maci-split-validators](https://github.com/macisamuele/jsonschema-rs/tree/maci-split-validators).
The branch is far from been really presentable (meant to provide a possible solution for #63)

Stranger6667:
Cool! Looking forward to seeing a pull request :)

Issue #80, opened 2020-05-22
Enum matching improvement


Stranger6667:
If all items in the enum are of the same type, then we can build a special validator for this type, in this case, we can speed up validation because for not matching types we won't need comparisons at all and for matching types, we'll have some improvement as well - we can store unwrapped values, unwrap the input value once during the matching step and then compare them in a loop, which should be faster.

```
{"enum": [1, 2, 3, 4, 5]}
```

If the input instance is `"foo"`, then we don't need to iterate over all variants, we can return `false` immediately. The trick will work only for homogenous arrays

PR #81, opened 2020-05-23
perf: Improve enum matching performance

Stranger6667:
PoC for #80 

Currently, for the `enum` bench, it gives 15% slower compilation, but 35-55% faster validation for valid numeric values and up to 50% improvement for validation of invalid values. It also gives some slight improvements for `small schema` bench, because of `enum: ["hello", "world"]` in it.

There are definitely possibilities to improve this PR

### DIFF
--- a/src/keywords/enum_.rs
+++ b/src/keywords/enum_.rs
@@ -1,9 +1,100 @@
 use super::{helpers, CompilationResult, Validate};
+use crate::primitive_type::{PrimitiveType, PrimitiveTypesBitMap};
 use crate::{
     compilation::{CompilationContext, JSONSchema},
     error::{error, no_error, CompilationError, ErrorIterator, ValidationError},
 };
 use serde_json::{Map, Value};
+
+pub struct NumberEnumValidator {
+    options: Value,
+    variants: Vec<f64>,
+}
+
+impl NumberEnumValidator {
+    #[inline]
+    pub(crate) fn compile(options: &Value, variants: Vec<f64>) -> CompilationResult {
+        Ok(Box::new(NumberEnumValidator {
+            options: options.clone(),
+            variants,
+        }))
+    }
+}
+
+impl Validate for NumberEnumValidator {
+    fn validate<'a>(&self, schema: &'a JSONSchema, instance: &'a Value) -> ErrorIterator<'a> {
+        if !self.is_valid(schema, instance) {
+            return error(ValidationError::enumeration(instance, &self.options));
+        }
+        no_error()
+    }
+
+    fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
+        if let Some(value) = instance.as_f64() {
+            self.variants.iter().any(|variant| *variant == value)
+        } else {
+            false
+        }
+    }
+
+    fn name(&self) -> String {
+        format!(
+            "enum: [{}]",
+            self.options
+                .as_array()
+                .expect("Always array")
+                .iter()
+                .map(|item| format!("{}", item))
+                .collect::<Vec<String>>()
+                .join(", ")
+        )
+    }
+}
+
+pub struct StringEnumValidator {
+    options: Value,
+    variants: Vec<String>,
+}
+
+impl StringEnumValidator {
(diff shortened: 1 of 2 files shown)


src/keywords/enum_.rs
```diff
@@ -1,10 +1,101 @@
...
+        Ok(Box::new(NumberEnumValidator {
+            options: options.clone(),
+            variants,
+        }))
+    }
+}
+
+impl Validate for NumberEnumValidator {
+    fn validate<'a>(&self, schema: &'a JSONSchema, instance: &'a Value) -> ErrorIterator<'a> {
+        if !self.is_valid(schema, instance) {
```
Stranger6667, line 26: We can avoid duplicating this method.

src/keywords/enum_.rs
```diff
@@ -1,10 +1,101 @@
...
+
+    fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
+        if let Some(value) = instance.as_f64() {
+            self.variants.iter().any(|variant| *variant == value)
+        } else {
+            false
+        }
+    }
+
+    fn name(&self) -> String {
```
Stranger6667, line 40: This method is also could be without duplication

src/keywords/enum_.rs
```diff
@@ -53,5 +144,36 @@ pub fn compile(
...
     _: &CompilationContext,
 ) -> Option<CompilationResult> {
-    Some(EnumValidator::compile(schema))
+    if let Value::Array(items) = schema {
+        let mut types = PrimitiveTypesBitMap::new();
+        let mut numbers = vec![];
+        let mut strings = vec![];
+        for item in items {
+            match item {
+                Value::Array(_) => types |= PrimitiveType::Array,
```
Stranger6667, line 153: I think it is possible to exit this loop somehow early if a different type occured

src/keywords/enum_.rs
```diff
@@ -53,5 +144,36 @@ pub fn compile(
     schema: &Value,
     _: &CompilationContext,
 ) -> Option<CompilationResult> {
-    Some(EnumValidator::compile(schema))
+    if let Value::Array(items) = schema {
+        let mut types = PrimitiveTypesBitMap::new();
+        let mut numbers = vec![];
```
Stranger6667, line 149: Probably we can avoid creating all these vectors

macisamuele:
I'm not entirely sure about this, the [specifications](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.1.2) mentions
> An instance validates successfully against this keyword if its value is equal to one of the elements in this keyword's array value.
> Elements in the array might be of any value, including null.

I don't see any specification of the values in the array should be of the same type and actually the tests provided by JSON-Schema-Test-Suite reflect this: https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft7/enum.json#L41

I would say that in order to improve enum validation performances we might do something like
```rust
// This is dummy code ... ;)
struct EnumValidator {
    array: Vec<Value>,
    string: Vec<String>,
    ....
}
```
for all the types and then during validation we do iterate on the "correct" array only according to the instance to validate ... this would save us some runtime  and be compliant with the specificatons

Stranger6667:
> I don't see any specification of the values in the array should be of the same type

You are right, I probably didn't reflect my intentions properly. The idea is that we can optimize check only for the case when all enum variants are of the same type and fallback to the current implementation otherwise so that the invariant stated in the spec still holds. I.e. if we have a string input value and we already know that there are no strings in the enum, then we can skip checking every element and immediately return false because all comparison operations will return false anyway because of a different type.

> or all the types and then during validation we do iterate on the "correct" array only according to the instance to validate

Indeed it should improve the performance in general, I'll explore this option as well! :)

macisamuele:
> Indeed it should improve the performance in general, I'll explore this option as well! :)

I mentioned that type of approach because if we're going to end merging #86 then for every instance to validate we don't even need to identify the type :). It might be a bit heavier in terms of bit in memory but I don't think that it would be a very high overhead

PR #83, opened 2020-05-23
$ref keyword caches built validators

macisamuele:
The goal of this PR is to reduce validation penalty performance while interacting with `$ref`s.

At the current state every time we're trying to validate such object we're actually re-building all the validators (which is essentially the compile part of the process).

I've tried to move validators build logic (currently in `RefValidator::ensure_validators`) into the compilation process, but it get's though due to potential circular dependencies with references.
In order to reduce the code changes and keep performances in consideration I'm opting for storing the validators behind a `RwLock`.
The decision of using it is mostly related to the fact that respect a `Mutex` it does have similar performances but at the same time allows multiple readers (in our case multiple parallel validations).

Performance improvements are over well 45% improvement for validation. Raw data in [here](https://gist.github.com/macisamuele/07ca443be23b62789e4b98d99cd928c8)

PS. once again this was spotted while working on [maci-split-validators](https://github.com/macisamuele/jsonschema-rs/tree/maci-split-validators) 😎

### DIFF
--- a/src/keywords/ref_.rs
+++ b/src/keywords/ref_.rs
@@ -1,60 +1,78 @@
 use super::{CompilationResult, Validate};
 use crate::{
     compilation::{compile_validators, CompilationContext, JSONSchema},
-    error::{error, ErrorIterator},
+    error::{error, ErrorIterator, ValidationError},
+    keywords::Validators,
 };
+use parking_lot::RwLock;
 use serde_json::Value;
 use url::Url;
 
 pub struct RefValidator {
     reference: Url,
+    /// Precomputed validators.
+    /// They are behind a RwLock as is not possible to compute them
+    /// at compile time without risking infinite loops of references
+    /// and at the same time during validation we iterate over shared
+    /// references (&self) and not owned references (&mut self).
+    validators: RwLock<Option<Validators>>,
 }
 
 impl RefValidator {
     #[inline]
     pub(crate) fn compile(reference: &str, context: &CompilationContext) -> CompilationResult {
         let reference = context.build_url(reference)?;
-        Ok(Box::new(RefValidator { reference }))
+        Ok(Box::new(RefValidator {
+            reference,
+            validators: RwLock::new(None),
+        }))
+    }
+
+    /// Ensure that validators are built and built once.
+    fn ensure_validators<'a>(&self, schema: &'a JSONSchema) -> Result<(), ValidationError<'a>> {
+        if self.validators.read().is_none() {
+            let (scope, resolved) =
+                schema
+                    .resolver
+                    .resolve_fragment(schema.draft, &self.reference, schema.schema)?;
+            let context = CompilationContext::new(scope, schema.draft);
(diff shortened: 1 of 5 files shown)


src/keywords/ref_.rs
```diff
@@ -1,60 +1,74 @@
...
+    error::{error, ErrorIterator, ValidationError},
+    keywords::Validators,
 };
 use serde_json::Value;
 use url::Url;
+use parking_lot::RwLock;
 
 pub struct RefValidator {
     reference: Url,
+    /// Precopmuted validators.
```
Stranger6667, line 13: `Precopmuted` -> `Precomputed`

src/keywords/ref_.rs
```diff
@@ -1,60 +1,74 @@
...
-                    ),
-                    Err(e) => error(e.into()),
-                }
-            }
-            Err(e) => error(e),
+        if let Err(err) = self.ensure_validators(schema) {
+            error(err)
+        } else {
+            Box::new(
+                self.validators.read().as_ref().expect("ensure_validators guarantees the presence of the validators")
```
Stranger6667: With having that I suspect that it might improve our current situation with cloning - now validators are moved from this scope to the `RefValidator` which means that we will not return anything owned by the current function and we can try to return references instead. What do you think? I am not completely sure about lifetimes

src/keywords/ref_.rs
```diff
@@ -1,60 +1,78 @@
...
 };
+use parking_lot::RwLock;
 use serde_json::Value;
 use url::Url;
 
 pub struct RefValidator {
     reference: Url,
+    /// Precopmuted validators.
+    /// They are behind a RwLock as is not possible to compute them
+    /// at compile time without risking infinite loops of references
```
Stranger6667, line 15: Do you mean the rust compilation or validator compilation? Could you please elaborate a bit here?

Stranger6667: Ah, I didn't completely read your comment in the PR

Stranger6667:
> -95.816% 

Just wow :)

macisamuele:
To be honest, in realistic cases (where references point to big schemas) I would expect even bigger improvements.
Think about Swagger specifications where the reference points to the schema of an endpoint response (ie. interacting with Kubernetes)

src/keywords/ref_.rs
macisamuele: The `RefValidator::validate` method needs to `collect` the `FlatMap` as `self.validators` is owned by the `RwLockGuard`. This has an impact on materialising a new `Vec` to contain the errors, but we're not cloning duplicating `ValidationError` instances.

I don't think that we might have ways around this (except if we decide to inject `unsafe`) which I would prefer not doing as performance improvements are already good enough.

We might, in the future, try to inspect all the cases where we have `...collect().into_iter()` an `ValidationError::into_owned` calls to try to remove them as they are indicators of creating memory just to make rust borrow-checker happy ;)

Stranger6667: :+1:

Issue #57, opened 2020-05-18
Provide explanation in errors when the input schema is not valid


Stranger6667:
At the moment it is only `Schema compilation error`

Stranger6667:
If we'll have a metaschema validation, then this will be covered. Details are in #28

nokome:
Thanks for `jsonschema` :pray:. I've just started using it, so perhaps a noob question...

With `0.6.0` I'm still only getting the error message `Schema compilation error` when using `JSONSchema::compile(&schema)` or `JSONSchema::options().with_meta_schemas().compile(&schema)`. Should I be doing something different, or does this require #100 to be completed?

Stranger6667:
Hi @nokome 

Sorry for the late reply! This problem went out of my scope and I think that it is separate - #100 will provide better error messages for sure, but at the moment meta schemas are only stored and not used for validation. I think that instead of `CompilationError` and `ErrorIterator` we can only have the latter and have it as the `Err` variant for the `JSONSchema::compile` return type.

Stranger6667:
Closed in favor of #198

Stranger6667:
@nokome

FYI - as #198 is merged and released, schema errors are reported in the same fashion as validation errors. There is an error message & `instance_path`  that may help with debugging

Issue #28, opened 2020-03-29
Store & use meta-schemas


Stranger6667:
If we'll validate the input schemas for conformance to the respective specs, then:

- We probably can skip a lot of our own checks during the compilation process
- There will be an understandable error message in case if the input schema is not valid

Regarding the implementation details - it can be done via `lazy_static!` so the schema is not re-compiled. In the perfect scenario I'd like to have it done via code-generation (like described here - #46)

darwin67:
It would be helpful to have better schema validation error messages.
We don't need code generation since the schemas are not pre-defined and are stored in a database. But without better error messages for schemas themselves, it's pretty hard for someone who might not be familiar with the specs to figure what they did wrong.

Stranger6667:
Sorry @darwin67 I missed your message.
The error messages should be covered by #100 - I hope to work on it in the next few months

PR #97, opened 2020-05-26
Add more lints

Stranger6667:


src/error.rs
```diff
@@ -14,7 +14,7 @@ use std::{
 ///
 /// It includes cases when during validation a reference is resolved into an invalid schema,
 /// which we can't know upfront because schemas can be in remote locations.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Copy, Clone)]
```
macisamuele, line 17: I'm not sure that `Copy` should be derived at the moment.

At the moment if we're validating with an invalid schema we're receiving a `CompilationError::SchemaError` instance which does not tell much (#57, #28).
In the long run I would expect that `CompilationError` does hold some information related to the schema that failed.

I'm mentioning this as the object might not be small and in such contexts is usually better to have the user be explicit about cloning data or passing references instead of having copies, of potentially big amount of data, done transparently.

Stranger6667: A fair point! Regarding the future - I was thinking about using metaschemas #28 so we can re-use `ValidationError` and it will provide more detailed information about the reason why the input schema is not valid. What do you think?

macisamuele: Yeah that's what I was expecting and that was the main driver for having concerns about `Copy`.
In my mind it was reading something among the lines _ValidationError is Copy_ and relating it with the struggling of trying to remove the `ValidationError::into_owner` method 😎

macisamuele: PS. Having it `Clone` is OK is more about I'm very reluctant to the idea of having `Copy` implemented except if I'm reasonable sure that the object will be very small.

Stranger6667: I think we can add `Clone` later when there will be progress with metaschemas

PR #98, opened 2020-05-27
fix: Precision loss in numeric validators

### LINKED ISSUE #84: Int to float conversion bug
In some numeric validators, we have a bug due to float conversion.`(1u64 << 54) - 1` is `18014398509481983` and `((2u64 << 53) - 1) as f64` is `18014398509481984.0` which is +1. This causes various bugs on the corner cases:

```{"minimum": 18014398509481984u64}``` should fail on `18014398509481984u64 - 1`, because the latter is smaller. but it doesn't because the latter rounds to `18014398509481984` and the validation fails. Interestingly that `18014398509481984u64 - 2` doesn't fail and after the power of  53 we have these ranges that should fail but they don't:
- 2 ^ 54 - 2 ^ 0 -> 2 ^ 54 - 1
- ...
- 2 ^ 63 - 2 ^ 9 -> 2 ^ 63 - 1

I.e with having the right part as the `minimum` value all values to the left part will pass validation

### PR DESCRIPTION
**Pull request by Stranger6667**
Related - #84 
This PR does not include a fix for `multipleOf`, since it is more complex than I initially though

Performance is about the same (even better up to 15% in some cases)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 
 - Wrong implementation of `is_valid` for `additionalProperties: false` keyword case. [#61](https://github.com/Stranger6667/jsonschema-rs/pull/61)
 - Possible panic due to type conversion in some numeric validators. [#72](https://github.com/Stranger6667/jsonschema-rs/pull/72)
+- Precision loss in `minimum`, `maximum`, `exclusiveMinimum` and `exclusiveMaximum` validators. [#84](https://github.com/Stranger6667/jsonschema-rs/issues/84)
 
 ## [0.2.0] - 2020-03-30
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/keywords/mod.rs [hunk 1]
```diff
@@ -260,4 +260,13 @@ mod tests {
         let compiled = JSONSchema::compile(&schema, None).unwrap();
         assert!(compiled.is_valid(&data))
     }
+
+    #[test_case(json!({"minimum": 18014398509481984u64}), json!(18014398509481984u64 - 1))]
+    #[test_case(json!({"maximum": 18014398509481984u64 - 1}), json!(18014398509481984u64))]
+    fn is_not_valid(schema: Value, instance: Value) {
```
macisamuele, line 266: I would suggest moving this tests into the correct keyword files.

By having this sort of unit-tests in the specific files allows you to have the test of the corner-cases closer to where you're implementing them


Not blocking ... is more about style :)

Stranger6667: :+1: Thanks!

Stranger6667:
I'll do `multipleOf` separately, probably using this approach - https://stackoverflow.com/a/62067066/1779619

Issue #110, opened 2020-06-10
Build wheels for other architectures


Stranger6667:
- MacOs
- Windows 32bit
- Windows 64bit

macisamuele:
If you don't mind I'd love to work on this as well as improving a bit the python infra (ie. Cargo.lock should be committed on the python project)

Stranger6667:
Absolutely! FYI I took a lot from this repo when was doing Python CI - https://github.com/huggingface/tokenizers

PR #111, opened 2020-06-11
Enhance python package and distribute wheels for multiple platforms

macisamuele:
This PR:
 * adds `jsonschema_rs.__build__` which contains [useful build information](https://github.com/PyO3/pyo3-built/) that will be relevant in case of support in the debugging of issues with the package
 * ensures that the python library build is reproducible: this is "the opposite" of #90 :) . Cargo.lock file should be present if we're building an artifact that won't be used by rust and a Python library is one example of it.
 * Removed bash scripts to build source and binary distributions. This is done in order to keep better support for windows developers (they might not have an easy access to bash) as well as we can leverage tox for dealing with Python virtual environments.
 * Better usage of github-actions: use `pypa/gh-action-pypi-publish` to publish artifacts on pypi

NOTE: In order to obtain reproducible builds we need to ensure that `python/Cargo.toml` contains a published version and not a path.
Unfortunately this might limit the capability to run the python tests against the `jsonschema` rust library modified. In order to workaround this I've modified the `setup.py` in order to accept in input an environmental variable (`UNRELEASED_JSONSCHEMA_PATH`) which should point to the root of `jsonschema-rs` git repo. By doing so we ensure that:
 * python build are reproducible at every commit
 * we can test the effect of the rust changes on the python library without actually releasing the rust library yet (and this is directly used in the CI)

I'm fully aware that this PR, again, is not small and I'm truly sorry for it :)

This fixes: #110 

cc: @Stranger6667


python/tox.ini
```diff
@@ -1,14 +1,32 @@
...
+  setuptools-rust
+commands =
+  python setup.py sdist
+
+[testenv:build-wheel]
+deps =
+  setuptools-rust
+  wheel
+commands =
+  python setup.py bdist_wheel
```
Stranger6667, line 32: Hm, it seems like the Linux wheels will not be manylinux compatible. Is it intentional? Or am I missing something?

macisamuele: As far as I can tell the generated wheels should be compatible with systems installing manylinux compatible.

I'll do some extra search and ensure that manylinux is supported.

Please note that I had to disable windows tests as they are failing somehow ;)

macisamuele: Today has been a very interesting learning around manylinux ... I was under the understanding that manylinux wheels are less preferable respect to `linux_x86_64` wheels as built with more "legacy" compatibilities but preferable due to the easier sharing and building.
Checking documentation online and pip implementation seems that the preference order is `manylinux2010_x86_64`, `manylinux1_x86_64` and then `linux`

The list is obtainable from
```
>>> from  pip._internal.models.target_python import get_supported
>>> get_supported()
```

I'm going to update the PR to build "regular" wheels for macos and windows, while opting for manylinux2010 wheels on linux

macisamuele: Using manylinux2014 wheels and re-enabled windows tests :)

Issue #112, opened 2020-06-12
Wheel publishing doesn't work on all platform


Stranger6667:
Probably we need to have a previous version of that step for mac :(

https://github.com/Stranger6667/jsonschema-rs/runs/766075069?check_suite_focus=true#step:7:9

cc @macisamuele

macisamuele:
@Stranger6667  thanks for pinging me on this.

This is ugly but I think that we can build all the wheels in parallel and then have a job that takes care only of publishing all the artifacts to pypi

Doing so would allow to have all-or-nothing published to pypi in case of transient failures which might be considered a benefit ;)

Stranger6667:
> This is ugly but I think that we can build all the wheels in parallel and then have a job that takes care only of publishing all the artifacts to pypi

Sounds good to me! :)

macisamuele:
One additional note https://github.com/Stranger6667/jsonschema-rs/runs/766075274?check_suite_focus=true

seems that `linux_x86_64` is not a platform accepted by pypi so we need to ensure that on linux we're publishing ONLY the manylinux wheel

Stranger6667:
Yeah ... the older wheel should be removed / excluded from uploading.

https://github.com/Stranger6667/jsonschema-rs/runs/766075274?check_suite_focus=true#step:9:11

Stranger6667:
Btw, when installing `sdist` package from pypi I faced with the following problem - `error: environment variable `OUT_DIR` not defined`

Should we have some fallback for such cases?

macisamuele:
> Yeah ... the older wheel should be removed / excluded from uploading.

This surprise me as `linux_x86_64` should be a valid tag. But it's not a problem to remove it.

> when installing sdist package ...

I'm very surprised by it.

At this point I would say that the latest release is not really installable, so I would suggest to [yank it](https://pypi.org/help/#yanked) just to avoid issues ;)

I'll try to get back to you with this fix ASAP (expect a few hours as I need to get a walk after the working day 😎)

macisamuele:
What I can say is that `OUT_DIR` variable is defined by the `build.rs` build script and is needed by `pyo3-built` to provide the metadata published in `jsonschema_rs.__build__`. 

Is that possible that it is not in the source binary?

I'm checking this idea

PR #113, opened 2020-06-12
Fix python source distribution and wheel uploads from all the platforms

macisamuele:
The artifact upload has been centralised into a single job to ensure that we're not partially publishing artifacts as well as simplification to have a single PyPi interaction point.

Additional changes of this PR are:
 * ensure that `build.rs` is present in the source distribution (addresses https://github.com/Stranger6667/jsonschema-rs/issues/112#issuecomment-643368821)
 * wheels should be uploadable on PyPi
    **NOTE**: They will also be available as build artifact in the github-actions interface

You can check the results on https://github.com/macisamuele/jsonschema-rs/runs/766206482 (the .github/workflows/python-release.yml wokflow file is the only modification respect the branch in this PR)

Fixes: #112

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -6,7 +6,7 @@ on:
       - python-v*
 
 jobs:
-  create_and_upload_macos_and_windows_wheels:
+  create_macos_and_windows_wheels:
     name: Wheels for Python ${{ matrix.python-version }} / ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


Stranger6667:
Thank you very much!

Re: yanking that release. I think since it wasn't installable I'll upload correct files instead of yanking

Stranger6667:
Oh, PyPi denies uploading the same filenames even if they are deleted. will issue 0.3.2

macisamuele:
@Stranger6667 just one detail, at the moment we publish only 64bits wheels. Please let me know if we would like to have the 32bits as well (even if I don't know how many might be using them)

Stranger6667:
@macisamuele 
I think it is ok to have only 64bits :)

macisamuele:
I guessed so, but just in case it would be enough to add the architecture into the github workflow.

This might get relevant once more arm based processors will be more spread into servers :)

PR #115, opened 2020-06-21
fix: `items` allows the presence of boolean schemas

macisamuele:
While working on https://github.com/macisamuele/jsonschema-equivalent I've noticed that having boolean schemas into `items` keyword result into an invalid schema, when instead it should be considered valid.

Example:
```json
{
    "additionalItems": false,
    "items": true
}
```
is considered an invalid schema but according to the [draft (section 6.4.1)](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.4.1)
> The value of "items" MUST be either a valid JSON Schema or an array of valid JSON Schemas.

and `true` or `false` are valid JSON Schemas.

The goal of this PR is to fix this.

PS: I'm aware that the example schema is unlikely to be composed by humans as it really is as restrictive as `true` schema, but it is still a possible valid schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - Enable Link-Time Optimizations and set `codegen-units` to 1. [#104](https://github.com/Stranger6667/jsonschema-rs/issues/104)
+- fix: `items` allows the presence of boolean schemas. [#115](https://github.com/Stranger6667/jsonschema-rs/pull/115)
 
 ## [0.3.0] - 2020-06-08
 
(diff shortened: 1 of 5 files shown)


Stranger6667:
Hi @macisamuele !

Thanks! Will release a new version today :)

macisamuele:
> Will release a new version today :)

Sure thing, in the mean time I'm trying to add more test coverage on the other project, so I can easily work with a patched crates.io ;) 

> Can you, please, add a changelog entry?

Doing it right now

Issue #101, opened 2020-06-09
Broken idn-hostname format on Draft 7


Stranger6667:
With the current test suite it gives:

```
---- draft7::optional_format_idn_hostname_0_11 stdout ----
thread 'draft7::optional_format_idn_hostname_0_11' panicked at 'Schema: {"format":"idn-hostname"}
Instance: "ःhello"
Error: It is supposed to be INVALID!', tests/test_suite.rs:5:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- draft7::optional_format_idn_hostname_0_6 stdout ----
thread 'draft7::optional_format_idn_hostname_0_6' panicked at 'Schema: {"format":"idn-hostname"}
Instance: "xn--X"
Error: It is supposed to be INVALID!', tests/test_suite.rs:5:1

---- draft7::optional_format_idn_hostname_0_7 stdout ----
thread 'draft7::optional_format_idn_hostname_0_7' panicked at 'Schema: {"format":"idn-hostname"}
Instance: "XN--aa---o47jg78q"
Error: It is supposed to be INVALID!', tests/test_suite.rs:5:1


failures:
    draft7::optional_format_idn_hostname_0_11
    draft7::optional_format_idn_hostname_0_6
    draft7::optional_format_idn_hostname_0_7
```

Issue #94, opened 2020-05-25
Refactor keywords into "basic" keywords and "composite" keywords


macisamuele:
The goal of this issue is to track the effort needed to create a well defined structure for:
 * "basic" keywords: keywords that should be validated in isolation
 * "composite" keywords": multiple keywords that can be validated together

At the current stage ([9767a9f921a85bc8295b0405f19ee64f48953b4c](https://github.com/Stranger6667/jsonschema-rs/tree/9767a9f921a85bc8295b0405f19ee64f48953b4c)) we already a single case of composite keywords:
 * `contentEncoding` and `contentMediaType`

The idea of creating such structure would allow us to:
 * extend the set of composite keywords as needed (#79)
 * keep better code separation between the keyword implementation and a "macro-keyword" one

Ideally the `compile_validators` should be looking like
```rust
let mut validators = Vec::with_capacity(object.len());
/* START NEW */
let mut object_keywords = object.keys().collect::<HashSet<_>>();
for (keywords, compilation_func) in get_all_composite_validators_for_draft(context.draft) {
    if ! object_keywords contains all keywords ! {
        if let Some(validator) = compilation_func(object, subschema, &context) {
                validators.push(validator?);
                ! Discard used keywords !
            }
        }
    }
}
/* END NEW */
for keyword in object_keywords { // object_keywords will contain the not used keywords
    if let Some(compilation_func) = context.draft.get_validator(keyword) {
        if let Some(validator) = compilation_func(object, object.get(keyword).unwrap(), &context) {
            validators.push(validator?)
        }
    }
}
Ok(validators)
```

This is just a rough idea, and I'm looking for feedbacks before trying something in this regard.
I think that having a better structure would allow us
 * to easily tackle #79 as hooking the logic into the code would be simpler
 * eventually allow us to enable/disable the composite keywords concept via cargo features or `JSONSchema::compile` flags

Stranger6667:
I'll try to provide some feedback tomorrow

Stranger6667:
[...]

PR #116, opened 2020-06-21
Bump json schema test suite

macisamuele:
The goal of this PR is to bump json-schema-test-suite.

NOTE: Tests of `resolver` needed to be updated in order to make them independent from the external testing suite (and so making easier future bumps)
I "just" copied the schema resulting from `load` without trying to simplify it to contain only the information useful for the test.


macisamuele:
I'll be fixing the tests soon ;)

macisamuele:
For tracking: #101 

@Stranger6667 : It will take me a bit longer than anticipated (especially if the idea is to fix the tests and not just ignore them).
I'll try to keep it scoped in time (as I was working on #94)

PR #118, opened 2020-06-22
Enable windows tests

macisamuele:
The goal of this PR is to ensure that all the tests are executed on Windows (either rust and python library).

In order to achieve this we need to make sure that the mock server is started by the same process running the tests (this is needed due to some Windows security limitations). I've decided that was easier to embed the mock-server instead of trying to workaround the Windows configs (even considering that I'm not very familiar with that platform 😎 )

Some context on the changes:
 * github workflows do not have the `Start background server` as *rust* and *python* will take care of spinning the mock server
    - on rust I'm using [json-schema-test-suite-rs](https://github.com/macisamuele/json-schema-test-suite-rs) procedural macro as it allows library owners to easily visualise their test (not hidden into the `draft` dependency) as well as it ensures that we run a mock-server with `mockito` directly in rust
    - on python I'm just running the flask mock-server as a pytest fixture in autouse (so we won't forget about adding it)
 * `UNRELEASED_JSONSCHEMA_PATH` *trick* has been removed: this is done because `cargo` knows how to deal with `path` and `version` defined at the same time (the rule of thumb is: path is used for local builds, version is used while publishing). Doing so helps us to get rid of a lot of "infra" code that require maintenance with no advantage (PS. Sorry as I did add that code on the first place)
 * `jsonschema` version is *strongly* pinned to a specific version (`= 0.3.1`).
    This is done in order to have test failures when the rust library is bumped without bumping the related dependency on the python project. Once again this is done to avoid forgetting to bump the `Cargo.lock` and the needed `jsonschema-rs` library for too long.
    The error will look like
```
maci:python/ (maci-enable-windows-tests✗) $ cargo build
error: failed to select a version for the requirement `jsonschema = "=0.3.2"`
  candidate versions found which didn't match: 0.3.1
  location searched: /Users/maci/github/jsonschema-rs
required by package `jsonschema-python v0.1.0 (/Users/maci/github/jsonschema-rs/python)`
consider running `cargo update` to update a path dependency's locked version
```

I hope that I've summarised all the important changes, if you think that I need to explain more in depth some details please ask ;)

cc: @Stranger6667


tests/test_suite.rs
```diff
@@ -1,12 +1,43 @@
...
+    let draft_version = match test_case.draft_version.as_ref() {
+        "draft4" => Draft::Draft4,
+        "draft6" => Draft::Draft6,
+        "draft7" => Draft::Draft7,
+        _ => panic!("Unsupported draft"),
+    };
+
+    let compiled = JSONSchema::compile(&test_case.schema, Some(draft_version)).unwrap();
+
+    let result = compiled.validate(&test_case.instance);
```
Stranger6667, line 21: It seems that checking equivalence of `compiled.validate` and `compiled.is_valid` is missing

macisamuele: I feel like ensuring that ensuring
 * if `schema.is_valid(instance)` then `schema.validate(instance)` will have no errors
 * if `!schema.is_valid(instance)` then `schema.validate(instance)` will have at least one error
should not be done into an integration test.

and so I did remove it as it felt like it was an unintentional addition, but this is a low-cost addition, so don't necessarily mind adding it back

Stranger6667: Hm, initially my thoughts were about keeping those shortcut implementations in sync with the more verbose ones and I felt like this is a convenient place to do so :)

PR #119, opened 2020-06-23
Combined validators setup + Initial impl for type string

macisamuele:
The goal of this PR is to define the concept of `combined` validators as presented on #94 .

This PR moves the combined handling of `contentEncoding` ad `contentMediaType` into the combined validators and defines some composed validators using `"type": "string"` constraint.

**NOTE**: The combined validators using type-string handle the `const` keyword as well
**WARNING**: The combined validators do report only the first/one `ValidationError` and not all the possible ones. I went this path mostly for simplicity.

This PR is published in order to do some more aggressive perf-test.
Mentioning this because I've noticed that:
 * no noticeable performance change by combining keywords
 * we might allow users of the lib to optimise the schema before validating (maybe via [`jsonschema-equivalent`](https://github.com/macisamuele/jsonschema-equivalent))
 * we might re-order keywords validation order (ie. validate type, const, ...) as it might allow more consistent and fast perf


macisamuele:
Rebased on latest master

Stranger6667:
Hi! I created the following benchmark:

```
bench!(
  name = "combined";
  schema = {"type": "string", "minLength": 5, "maxLength": 10};
  valid = "foobar";
  invalid = "foo";
);
```

And on this branch I have the following results:

```
jsonschema-rs combined compile typestringminLength5maxLength10
                        time:   [480.66 ns 481.62 ns 482.58 ns]
                        change: [+229.46% +231.59% +234.06%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe

jsonschema-rs combined is_valid valid foobar
                        time:   [5.5320 ns 5.5401 ns 5.5495 ns]
                        change: [-49.408% -49.191% -48.973%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild
  3 (3.00%) high severe

jsonschema-rs combined validate valid foobar
                        time:   [18.553 ns 18.565 ns 18.580 ns]
                        change: [-32.724% -32.525% -32.271%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe

jsonschema-rs combined is_valid invalid foo
                        time:   [3.8436 ns 3.8465 ns 3.8498 ns]
                        change: [-40.876% -40.608% -40.288%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe

jsonschema-rs combined validate invalid foo
                        time:   [80.710 ns 80.787 ns 80.871 ns]
                        change: [-7.3770% -7.0675% -6.7518%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
```

This seems nice from the validation point of view (especially ~50% improvement on the valid input via `is_valid`), but for some reason, the compilation extremely slowed down, but probably it can be improved

> no noticeable performance change by combining keywords

Hm, it differs from my observations. What was your testing data?

macisamuele:
@Stranger6667 I don't really have trace anymore of the data used for the benchmark ;( 

I do realise that schema compilation will be slowed down by this ... clearly we can do something to reduce the impact but I do still expect a performance drop in `JSONSchema::compile` but I was not expecting this big (twice as long).

I'll try to check what we can do to speedup compilation a bit, probably having sets of unprocessed keys (that require a string clone) is creating the perf hit (or a good chunk of it)

macisamuele:
Going to close this review as I'm struggling to find bandwidth to identify the performance penalty added at compile time.

I'll try to revive this and/or gate this with options such that users that use the JSONSchema into a long running service might benefit from faster validation at the cost of some additional initialisation time.

PR #122, opened 2020-06-29
fix: Add `idna` dependency to improve idn-hostname format correctness

macisamuele:
The goal of this PR, as visible on the description, is to improve the correctness of `idn-hostname` format.

Checking around I found no perfect rust library that does this validation and implementing the whole RFC seems really out of scope.
Among the different available libraries (idna, unic-idna, unic-idna-punicode) `idna` seems the most maintained and accurate (allows us to remove two of the three optional ignored tests).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ reqwest = { version = ">= 0.10", features = ["blocking", "json"]}
 parking_lot = ">= 0.1"
 num-cmp = ">= 0.1"
 paste = ">= 0.1"
+idna = ">= 0.2"
 
 [dev-dependencies]
 criterion = ">= 0.1"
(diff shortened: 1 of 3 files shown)


Stranger6667:
Cool! Could you please enable relevant Python tests as well?

Stranger6667:
Hm, currently we ignore whole file there, probably we can make it separately then

macisamuele:
@Stranger6667 if we try to bump again the test-suite we will find more failures.
I'm currently releasing an update on the proc-macro for the tests such that we can use regexes.

I'll be updating the relevant python tests as well as updating the readme mentioning that `idn-hostname` optional tests are currently ignored on draft7

macisamuele:
@Stranger6667 I think that your input on this question is extremely valuable.
 
Bumping the json-schema test suite causes 17 tests (`optional_format_idn_hostname_0_\d+`) failures.
Adding the `idna` dependency brings us to 14 test failures. So the dependency is able to catch only 3 cases.

Do you think that is worth adding it as seems being so in-accurate?

Stranger6667:
I think it worth adding. The dependency is quite small and we might benefit from improvements on the `idna` side in the future even though it doesn't fix may test cases at the moment. For now, I agree that we can just ignore all idn-hostname test cases.

macisamuele:
Cool I'll be publishing a new pr ignoring the tests

PR #129, opened 2020-07-05
Customizable media type and content encoding

macisamuele:
Publishing this as draft in order to gather feedback on the approach.
**NOTE**: Currently documentation of public methods is missing as well as tests on usage of custom encoding.

The general idea is to allow customisation of the `JSONSchema::compile` behaviour via `JSONSchemaCompileConfig`.

At the current state I'm extending it to allow customisable definition of `mediaType` and `contentEncoding` (as we're only supporting JSON and Base64) but it might allow (something that would eventually be provided in subsequent PRs);
 * custom formats (ie. what if we would like to honor user defined formats) (ie. `idn-hostname` is not properly supported in rust, but we might embed the `python` library for the python binding)
 * embedding schema pre-processing/optimisation performed before the schema compilation (ie. embedding jsonschema-equivalent)

NOTE: In order to achieve it by ensuring that we're limiting clone (for example in `$ref` keyword) we need to have a specific lifetime into `Validate` trait, but overall I feel like it was an acceptable tradeoff to have.

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -44,19 +44,15 @@ impl From<JSONSchemaError> for PyErr {
     }
 }
 
-fn get_draft(draft: Option<u8>) -> PyResult<Draft> {
-    if let Some(value) = draft {
-        match value {
-            DRAFT4 => Ok(jsonschema::Draft::Draft4),
-            DRAFT6 => Ok(jsonschema::Draft::Draft6),
-            DRAFT7 => Ok(jsonschema::Draft::Draft7),
-            _ => Err(exceptions::ValueError::py_err(format!(
-                "Unknown draft: {}",
-                value
-            ))),
-        }
-    } else {
-        Ok(jsonschema::Draft::default())
+fn get_draft(draft: u8) -> PyResult<Draft> {
+    match draft {
+        DRAFT4 => Ok(jsonschema::Draft::Draft4),
+        DRAFT6 => Ok(jsonschema::Draft::Draft6),
+        DRAFT7 => Ok(jsonschema::Draft::Draft7),
+        _ => Err(exceptions::ValueError::py_err(format!(
+            "Unknown draft: {}",
+            draft
+        ))),
     }
 }
 
@@ -72,11 +68,14 @@ fn get_draft(draft: Option<u8>) -> PyResult<Draft> {
 #[pyfunction]
 #[text_signature = "(schema, instance, draft=None)"]
 fn is_valid(schema: &PyAny, instance: &PyAny, draft: Option<u8>) -> PyResult<bool> {
-    let draft = get_draft(draft).map(Some)?;
     let schema = ser::to_value(schema)?;
     let instance = ser::to_value(instance)?;
-    let compiled =
-        jsonschema::JSONSchema::compile(&schema, draft).map_err(JSONSchemaError::Compilation)?;
+    let mut config = jsonschema::CompilationConfig::default();
+    if let Some(raw_draft_version) = draft {
+        config.set_draft(get_draft(raw_draft_version)?);
+    }
+    let compiled = jsonschema::JSONSchema::compile(&schema, Some(config))
+        .map_err(JSONSchemaError::Compilation)?;
     Ok(compiled.is_valid(&instance))
 }
 
@@ -100,13 +99,16 @@ struct JSONSchema {
 impl JSONSchema {
     #[new]
     fn new(schema: &PyAny, draft: Option<u8>) -> PyResult<Self> {
-        let draft = get_draft(draft).map(Some)?;
         let raw_schema = ser::to_value(schema)?;
+        let mut config = jsonschema::CompilationConfig::default();
+        if let Some(raw_draft_version) = draft {
+            config.set_draft(get_draft(raw_draft_version)?);
+        }
         // Currently, it is the simplest way to pass a reference to `JSONSchema`
         // It is cleaned up in the `Drop` implementation
         let schema: &'static Value = Box::leak(Box::new(raw_schema));
         Ok(JSONSchema {
-            schema: jsonschema::JSONSchema::compile(schema, draft)
+            schema: jsonschema::JSONSchema::compile(schema, Some(config))
                 .map_err(JSONSchemaError::Compilation)?,
             raw_schema: schema,
         })
--- a/src/compilation/config.rs
+++ b/src/compilation/config.rs
@@ -1,10 +1,15 @@
-use crate::schemas;
+use crate::{
+    content_media_type::{ContentMediaTypeCheckType, DEFAULT_CONTENT_MEDIA_TYPE_CHECKS},
+    schemas,
+};
 use serde_json::Value;
+use std::{collections::HashMap, fmt};
 
 #[allow(missing_docs)]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Default)]
 pub struct CompilationConfig {
     pub(crate) draft: Option<schemas::Draft>,
+    content_media_type_checks: HashMap<String, Option<ContentMediaTypeCheckType>>,
 }
 
 #[allow(missing_docs)]
@@ -24,5 +29,40 @@ impl CompilationConfig {
     pub fn set_draft(&mut self, draft: schemas::Draft) -> &mut Self {
         self.draft = Some(draft);
         self
+    }
+
+    pub(crate) fn content_media_type_check(
+        &self,
+        media_type: &str,
+    ) -> Option<ContentMediaTypeCheckType> {
+        if let Some(value) = self.content_media_type_checks.get(media_type) {
+            *value
+        } else if let Some(value) = DEFAULT_CONTENT_MEDIA_TYPE_CHECKS.get(media_type) {
+            Some(*value)
+        } else {
+            None
+        }
+    }
+
+    pub fn add_content_media_type_check<IS: Into<String>>(
+        &mut self,
+        media_type: IS,
+        media_type_check: Option<ContentMediaTypeCheckType>,
+    ) -> &mut Self {
+        self.content_media_type_checks
+            .insert(media_type.into(), media_type_check);
+        self
+    }
+}
+
+impl fmt::Debug for CompilationConfig {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.debug_struct("CompilationConfig")
+            .field("draft", &self.draft)
+            .field(
+                "content_media_type_checks",
+                &self.content_media_type_checks.keys(),
+            )
+            .finish()
     }
 }
--- a/src/compilation/config.rs
+++ b/src/compilation/config.rs
@@ -1,4 +1,8 @@
 use crate::{
+    content_encoding::{
+        ContentTypeCheckType, ContentTypeConverterType, DEFAULT_CONTENT_ENCODING_CHECKS,
+        DEFAULT_CONTENT_ENCODING_CONVERTERS,
+    },
     content_media_type::{ContentMediaTypeCheckType, DEFAULT_CONTENT_MEDIA_TYPE_CHECKS},
     schemas,
 };
@@ -10,6 +14,8 @@ use std::{collections::HashMap, fmt};
 pub struct CompilationConfig {
     pub(crate) draft: Option<schemas::Draft>,
     content_media_type_checks: HashMap<String, Option<ContentMediaTypeCheckType>>,
+    content_encoding_checks: HashMap<String, Option<ContentTypeCheckType>>,
+    content_encoding_converters: HashMap<String, Option<ContentTypeConverterType>>,
 }
 
 #[allow(missing_docs)]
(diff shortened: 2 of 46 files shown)


python/src/lib.rs
```diff
@@ -72,11 +68,14 @@ fn get_draft(draft: Option<u8>) -> PyResult<Draft> {
 #[pyfunction]
 #[text_signature = "(schema, instance, draft=None)"]
 fn is_valid(schema: &PyAny, instance: &PyAny, draft: Option<u8>) -> PyResult<bool> {
-    let draft = get_draft(draft).map(Some)?;
     let schema = ser::to_value(schema)?;
     let instance = ser::to_value(instance)?;
-    let compiled =
-        jsonschema::JSONSchema::compile(&schema, draft).map_err(JSONSchemaError::Compilation)?;
+    let mut config = jsonschema::JSONSchemaCompileConfig::default();
```
Stranger6667, line 73: I think that instead of calling `config.set_draft` it is cleaner to detect draft + create a config object, which is 2 steps when the current implementation is 3 steps - create a config, detect draft, set draft

src/compilation/config.rs
```diff
@@ -0,0 +1,128 @@
...
+    },
+    media_type::{MediaTypeCheckType, DEFAULT_MEDIA_TYPE_CHECK},
+    schemas,
+};
+use serde_json::Value;
+use std::{collections::HashMap, fmt};
+
+#[allow(missing_docs)]
+#[derive(Clone, Default)]
+pub struct JSONSchemaCompileConfig {
```
Stranger6667, line 14: What do you think about `CompilationConfig` instead?

src/compilation/config.rs
```diff
@@ -0,0 +1,128 @@
...
+    draft: Option<schemas::Draft>,
+    custom_content_media_type_checks: HashMap<String, Option<MediaTypeCheckType>>,
```
Stranger6667, line 16: I think `custom_` prefix is not necessary

src/compilation/config.rs
```diff
@@ -0,0 +1,128 @@
...
+    custom_content_encoding_checks: HashMap<String, Option<ContentTypeCheckType>>,
+    custom_content_encoding_converters: HashMap<String, Option<ContentTypeConverterType>>,
+}
+
+#[allow(missing_docs)]
+impl JSONSchemaCompileConfig {
+    pub(crate) fn draft(&self) -> schemas::Draft {
+        self.draft
+            .expect("JSONSchema::compile should have defined a specific draft version.")
```
Stranger6667, line 25: Not sure if it should be failable at all... e.g. when is it ok to panic? We always have a default - Draft 7 at the moment. Having an infallible data layout will help us to avoid extra actions like `set_draft_if_missing` which seems unnecessary to me

src/compilation/config.rs
```diff
@@ -0,0 +1,128 @@
...
+};
+use serde_json::Value;
+use std::{collections::HashMap, fmt};
+
+#[allow(missing_docs)]
+#[derive(Clone, Default)]
+pub struct JSONSchemaCompileConfig {
+    draft: Option<schemas::Draft>,
+    custom_content_media_type_checks: HashMap<String, Option<MediaTypeCheckType>>,
+    custom_content_encoding_checks: HashMap<String, Option<ContentTypeCheckType>>,
```
Stranger6667, line 17: What is the case for having `None` as a possible value in these hashmaps?

src/compilation/mod.rs
```diff
@@ -31,31 +36,37 @@ impl<'a> JSONSchema<'a> {
...
+        let mut jsonschema = JSONSchema {
             schema,
             resolver,
-            validators,
-        })
+            validators: Vec::with_capacity(0),
+            context,
+        };
+
+        let mut validators = compile_validators(schema, &jsonschema.context)?;
```
Stranger6667, line 65: It seems more clear to create `JSONSchema` in the end, rather than modifying it after creation, especially for this case - it is unconditionally replaced

src/compilation/mod.rs
```diff
@@ -31,31 +36,37 @@ impl<'a> JSONSchema<'a> {
...
         // Draft is detected in the following precedence order:
         //   - Explicitly specified;
         //   - $schema field in the document;
         //   - Draft7;
-        let draft = draft.unwrap_or_else(|| {
-            schemas::draft_from_schema(schema).unwrap_or(schemas::Draft::Draft7)
-        });
+
+        let mut config = config.unwrap_or_default();
+        config.set_draft_if_missing(schema);
```
Stranger6667, line 47: Maybe it will be better to disallow creating `JSONSchemaCompileConfig` without `draft`

src/compilation/mod.rs
```diff
@@ -1,26 +1,31 @@
...
 
 pub const DEFAULT_ROOT_URL: &str = "json-schema:///";
 
 /// The structure that holds a JSON Schema compiled into a validation tree
 #[derive(Debug)]
 pub struct JSONSchema<'a> {
-    pub(crate) draft: schemas::Draft,
     pub(crate) schema: &'a Value,
-    pub(crate) validators: Validators,
+    pub(crate) validators: Validators<'static>,
```
Stranger6667, line 26: Why is it `'static` ? It doesn't seem to me, that anything inside `Validators` should have a static lifetime

src/compilation/mod.rs
```diff
@@ -85,59 +96,12 @@ impl<'a> JSONSchema<'a> {
...
-    /// Build a new URL. Used for `ref` compilation to keep their full paths.
-    pub(crate) fn build_url(&self, reference: &str) -> Result<Url, ParseError> {
-        Url::options().base_url(Some(&self.scope)).parse(reference)
-    }
-}
-
 /// Compile JSON schema into a tree of validators.
 #[inline]
-pub fn compile_validators(
+pub fn compile_validators<'context>(
```
Stranger6667, line 101: Hm, shouldn't `context` variable have this lifetime as well?

src/content_encoding.rs
```diff
@@ -0,0 +1,38 @@
...
+    instance_string: &str,
+) -> Result<String, ValidationError<'a>> {
+    match base64::decode(instance_string) {
+        Ok(value) => Ok(String::from_utf8(value)?),
+        Err(_) => Err(ValidationError::format(instance, "base64")),
+    }
+}
+
+lazy_static::lazy_static! {
+    pub(crate) static ref DEFAULT_CONTENT_ENCODING_CHECK: HashMap<&'static str, ContentTypeCheckType> = {
```
Stranger6667, line 27: Hm, maybe `DEFAULT_CONTENT_ENCODING_CHECKS`? to reflect that there could be many checks

src/content_encoding.rs
```diff
@@ -0,0 +1,38 @@
...
+        let mut map: HashMap<&'static str, ContentTypeCheckType> = HashMap::with_capacity(1);
+        let _ = map.insert("base64", is_base64);
+        map
+    };
+
+    pub(crate) static ref DEFAULT_CONTENT_ENCODING_CONVERT: HashMap<&'static str, ContentTypeConverterType> = {
```
Stranger6667, line 33: Maybe `DEFAULT_CONTENT_ENCODING_CONVERTERS`? It seems more in line with the type name

src/content_encoding.rs
```diff
@@ -0,0 +1,38 @@
...
+        let mut map: HashMap<&'static str, ContentTypeConverterType> = HashMap::with_capacity(1);
+        let _ = map.insert("base64", from_base64);
```
Stranger6667, line 35: To we need `_` variable?

src/validator.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
 use serde_json::{Map, Value};
 use std::fmt;
 
-pub trait Validate: Send + Sync + ToString {
+pub trait Validate<'context>: Send + Sync + ToString {
```
Stranger6667, line 8: > NOTE: In order to achieve it by ensuring that we're limiting clone (for example in $ref keyword) we need to have a specific lifetime into Validate trait, but overall I feel like it was an acceptable tradeoff to have.

Could you, please, elaborate on why exactly do we need this lifetime? Not sure if I completely understand what clones you mean

macisamuele:
> Thanks! I like the idea :) It would be great to allow the users to customize this behavior :) However, I think we need to iterate a bit on the implementation and data structures

That was exactly the reason of the draft ... I pulled out an idea and a dummy iteration in order to gather feedback and possible improvements before stabilising the implementation.


**NOTE**: I had this in mind because I'd like to allow to plug in optional schema processors (which might require some changes on how we store the raw-schema, but I have no dummy code to show yet) or allow customisable format validators ;)

src/compilation/config.rs
macisamuele: This would require some documentation ... and I'm sorry that I've not done it yet.

The idea is that whenever we're getting a a content encoding check we're giving priority to the "user-defined" one first, if we haven't found it then we use the defaults.

The idea of allowing `None` in there is to allow users to "de-register" default formats.

Using this approach in order to have a static and immutable representation of the default handled mediaTypes, contentEncoding (and in the future formats) to reduce as much as possible the memory impact in case many many JSONSchema instances (or configs) are created

src/compilation/config.rs
macisamuele: That's correct, but this is partially needed to allow users to not specify the Draft version when `JSONSchema` is created (or in `jsonschema::is_valid`).

In compilation phase we would need to allow the user to specify or not specify the draft in the configuration.
If the draft is not configured (set to `None`) then we try to detect it from the schema and/or default it to `Draft7` (this is what `set_draft_if_missing` does).

We might be looking for an easier approach or just avoid the panic here by defaulting to`Draft::default()`.

My personal take is that an `expect` here would be acceptable because the invariance is guaranteed by the library and not by user-code and tests would quickly catch this issue. Using `Draft::default()` here would potentially hide the issue and lead to a "broken" release (where draft7 is used if the user has not specified the draft version)

src/compilation/config.rs
macisamuele: I agree ... as there was no doc I preferred the redundancy for now so at least it passed the message of what we should store in the field 😎

src/compilation/config.rs
macisamuele: Totally in for this ... the proposed name is far from being communicative and is unreasonably long.

python/src/lib.rs
macisamuele: I'm not sure I see your point in here ... can you post a snippet of what you have in mind?

macisamuele: I'm not sure I see your point in here ... can you post a snippet of what you have in mind?

src/validator.rs
macisamuele: Currently `RefValidator` builds the `validators` of the reference only when invoked for validation (and we have the `RWLock` to ensure that eventually we would compile it only once) ([here](https://github.com/Stranger6667/jsonschema-rs/blob/master/src/keywords/ref_.rs#L39-L40)).
This means that we need to preserve the config information within the `RefValidator` instance.

Let's assume that we don't have a lifetime, then we need to clone it, instead by enforcing the lifetime we can just store it via a CoW instance.

We might get rid of it if we can pull the build config from the JSONSchema passed along during validation, but I did not attempt it as I was looking for a general feedback on the idea first :)

Definitely if we can remove the lifetimes and ensuring that we're not cloning info then it would be preferrable

src/content_encoding.rs
macisamuele: I'm usually used to have [`unused-results`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-results) lint enable to prevent myself from forgetting to use the result of a method. This means that I do need to assign the result to something and `_` is the usual "trash-bin".

Definitely this boilerplate will be removed into the more final solution (as in this repo the lint is not enforced)

src/compilation/mod.rs
macisamuele: Yes, that's correct I was experimenting with something related to lifetimes and I ended up keeping in the repo instead of discarding the change

src/validator.rs
macisamuele: Cross-checked again and I don't know why I kept it ... we actually don't need it (at least not for now).

So I'm going to remove it and keep this PR scoped

python/src/lib.rs
Stranger6667: I had in mind the code it was before, but after the latest changes it is exactly what I meant:

```rust
Ok(JSONSchema {
            schema,
            resolver,
            validators,
            context,
        })
```

Stranger6667: Oh, it seems I confused this one with the lib itself, but it is actually Python bindings. The idea was to create `JSONSchemaCompileConfig` with draft instead of creating a `mut` one and updating it

src/compilation/config.rs
Stranger6667: I think that having ``pub(crate) draft: schemas::Draft`` is better, because we statically don't have a state where it is `None` and we don't need tests for it or using `expect` at all. If the user doesn't want to specify a draft, then they can rely on `Default::default()`. 

> Using Draft::default() here would potentially hide the issue and lead to a "broken" release (where draft7 is used if the user has not specified the draft version)

Hm, this is the current behavior, right? I mean it seems expected to me if I don't specify a draft version / config completely, then I let the library select it for me. 

Maybe we can use some kind of builder pattern instead.

```rust
let config = CompilationConfig::options().draft(Draft7).build();  // Or other naming - but some kind of builder pattern
let compiled = JSONSchema::compile(schema, Some(config));
```

Or in the `JSONSchema` itself (like it is in `url::Url` )

```rust
let compiled = JSONSchema::options().draft(Draft7).compile(schema)?;
```

What do you think?

macisamuele: > Hm, this is the current behavior, right?

Not quite, the current implementation (either master or this PR) does try to detect the version from the input schema and then if still no draft is defined we use `Draft7`.
Let's use the following schema as example
```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    ...
}
```
If we do compile this without defining the draft the schema should be compiled with `Draft6`, but if we do use `Draft::default()` it would use `Draft7`.

So we still need to preserve the information of "was draft defined by the user?"

---

About using the `JSONSchema::options`, it might be a clever idea as it would remove the naming difficulty and adhere to a pattern that seems getting more and more traction in the rust community.

I might try to make some progress on this maybe over the weekend though :)

Stranger6667: Yep, I meant using `default()` as the last option in that chain (how it is actually done now, but with explicitly using `Default::default()`)

> About using the JSONSchema::options, it might be a clever idea as it would remove the naming difficulty and adhere to a pattern that seems getting more and more traction in the rust community.

> I might try to make some progress on this maybe over the weekend though :)

I think it might be cool to have it as a separate PR to reduce the affected surface :) and then rebase

And, Samuele, thank you very much for working on this! :)

python/src/lib.rs
macisamuele: According to the current `JSONSchemaCompileConfig` implementation we still need two lines to do so ... I'm not going to focus on this for now as on a different open comment we're evaluating the idea of using `::options()`.

Keeping this open to ensure that we don't forget about it

src/compilation/config.rs
macisamuele: > I think it might be cool to have it as a separate PR

I agree ... over the weekend I'll try to create a PR where options are defined and they will have only the `Draft` handling (so it will small and nicer) ... after that other PRs will be spawned to add contentMediaType and contentMediaEncoding.

The idea of having them over here (discussion PR) was to show why I was thinking to enable a similar flow.
Sometimes "seeing" the effect is more effective than a very long and abstract description 😎

macisamuele:
Closing this draft PR as #131 and #133 have implemented the presented idea and they are now on master

PR #132, opened 2020-07-23
chore: Enforce clippy checks on testing code as well

macisamuele:
Update github actions to use [`clippy-check`](https://github.com/actions-rs/clippy-check) in order to have nice PR reports as well
Clippy rules are a bit more related for testing code (as we want to allow, for example, to use `Option::unwrap` API)


Stranger6667:
Thanks!

PR #133, opened 2020-07-23
Extend json schema options with content media type and content encoding

macisamuele:
The goal of this PR is to complete the feature proposal drafted in #129 

In order to simplify external crates to define methods for content encoding/media type support I've modified the signature such that `ValidationError` instances are created only within `jsonschema` crate instead of returned by the check/converter method.
This allows us to not do much work on exposing the all APIs of `ValidationError` (especially considering that it might change in the future).

I don't expect specific performance regressions, but I did not really tested them :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `ToString` trait implementation for validators.
 - Define `JSONSchema::options` to customise `JSONSchema` compilation [#131](https://github.com/Stranger6667/jsonschema-rs/issues/131)
+- Allow user-defined `contentEncoding` and `contentMediaType` keywords
 
 ### Fixed
 
--- a/src/keywords/content.rs
+++ b/src/keywords/content.rs
@@ -1,24 +1,22 @@
 //! Validators for `contentMediaType` and `contentEncoding` keywords.
 use crate::{
     compilation::{context::CompilationContext, JSONSchema},
+    content_media_type::ContentMediaTypeCheckType,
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - `ToString` trait implementation for validators.
 - Define `JSONSchema::options` to customise `JSONSchema` compilation [#131](https://github.com/Stranger6667/jsonschema-rs/issues/131)
+- Allow user-defined `contentEncoding` and `contentMediaType` keyworkds
```
Stranger6667, line 9:
```suggestion
- Allow user-defined `contentEncoding` and `contentMediaType` keywords
```

src/keywords/content.rs
```diff
@@ -5,20 +5,17 @@ use crate::{
...
     validator::Validate,
 };
-use serde_json::{from_str, Map, Value};
+use serde_json::{Map, Value};
 
 /// Validator for `contentMediaType` keyword.
 pub(crate) struct ContentMediaTypeValidator {
     media_type: String,
-    func: for<'a> fn(&'a Value, &str) -> ErrorIterator<'a>,
+    func: fn(&str) -> bool,
```
Stranger6667, line 13: What do you think about using `ContentMediaTypeCheckType` here and maybe below when applicable?

macisamuele: Totally agree ... I used the lazy `cargo fix` approach ;)

PR #134, opened 2020-07-25
Reduce explicit returns

macisamuele:
This PR is mostly introducing stylistic changes aimed to have the code-base reflecting the common rust style.

```rust
if foo() {
    return something;
}
return something_else;
```
can be written with a more idiomatic format
```rust
if foo() {
    something
} else {
    something_else
}
```

Few additional changes are:
 * `Validate::validate` relies on the result of `Validate::is_valid`: This allows to reduce code-duplication and should help in terms of binary size
 * Usage of iterators instead of `for` loops with inner results: This would allow for the same performance but with a more idiomatic style and would help in tuning more performance with parallel iterations if we can prove that this is beneficial (check contains and required files)

### DIFF
--- a/src/keywords/additional_properties.rs
+++ b/src/keywords/additional_properties.rs
@@ -123,11 +123,12 @@ impl AdditionalPropertiesNotEmptyFalseValidator {
     #[inline]
     pub(crate) fn compile(properties: &Value) -> CompilationResult {
         if let Value::Object(properties) = properties {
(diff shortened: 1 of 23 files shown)


Stranger6667:
Cool! Seems nice to me! Could you, please, fix the `rustfmt` build?

macisamuele:
@Stranger6667 ... sorry forgot to run it ;)

Issue #144, opened 2020-11-10
Include path/key in error message ?


yohanboniface:
For example:

```python
import jsonschema_rs

validator = jsonschema_rs.JSONSchema({"type": "object", "properties": {"foo": {"type": "integer"}}})
validator.validate({"foo": None})
```

This will raise an error with the string `"'null' is not of type 'integer'"`. With a more complex schema, this message is not really helpful to understand what's wrong in the data. Is there a way to retrieve the property that hold that value ? Or maybe better a path to this property ?

I imagine something like this: `"path.to.foo: 'null' is not of type 'integer'"`

And it would be even better if the error instance would contain that path in a dedicated attribute, so one can format the message easily (say for example to return it in json).

Thanks :)

Stranger6667:
Hi!

Indeed, it will be helpful! I think that #100  should cover it; however, I don't know when I will have the time for it :( PRs are welcome, though.

yohanboniface:
For the record, another message that could be clearer:

```python
import jsonschema_rs

validator = jsonschema_rs.JSONSchema({"type": "object", "properties": {"foo": {"type": "integer"}}, "additionalProperties": False})
validator.validate({"bar": "baz"})
```

Error message will be `False schema does not allow '"bar"'`.

wmain:
Big +1 to this issue. Without paths errors are essentially useless on schemas/data of our size.

Stranger6667:
Hi, long overdue, but here are some updates for the most recent Rust version (changes were not released for Python bindings yet):

- Improved error message for the `additionalProperties` keyword: `Additional properties are not allowed ('bar' was unexpected)` which is similar to what Python's `jsonschema` is using;
- The `ValidationError` struct now has the `instance_path` attribute that could be treated as a JSON Pointer that erroneous part of the validated instance;

There are some complications on the Python side, though, as `PyO3` does not provide a way to customize exception types :( But they are solvable, and I'll post updates on that matter here.

Stranger6667:
Here are few updates.

### New error messages include instance paths:

```
null is not of type "integer"

On instance["foo"]:
    null
```

Which is similar to what `jsonschema` is doing.

### Schema paths

Once #199 is ready, I'll expand the current error messages with schema paths.

### Separate attributes

Once `PyO3==0.14` is released (which should happen pretty soon), I'll add separate attributes for `instance_path` and `schema_path` (if it will be ready at that point).

Stranger6667:
[...]

Issue #100, opened 2020-06-07
Implement output formatting from Draft 2019-09


Stranger6667:
https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10

It does seem like a good way to standardize error output, regardless of the used schema version.

- [x] [Flag](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10.4.1)
- [x] [Basic](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10.4.2)
- [ ] [Detailed](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10.4.3)
- [ ] [Verbose](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10.4.4)

driv3r:
Hey @Stranger6667 ! I wanted to look into this issue, any up front pointers would be awesome.

Stranger6667:
Hi @driv3r !

Thank you for taking a step :) My initial plan was:

- store paths inside validators (as `String` or `Vec<String>`, not sure which one is more convenient). So, each validator knows its location within the schema.
- have a pointer (`Vec<String>` - JSON Pointer chunks) to a part of data being validated. This one could not be necessary in all cases, as we often can derive it from the current validator path. Skipping it for such cases seems like optimization to me and can be done later. 
- If an error happens, then `ValidationError` might take the validator's path & pointer to the current data part
- Implement a function on `ErrorIterator` (probably requires making it a separate struct & implementing `IntoIterator` on it) that will take some kind of "format" enum value that will consume the iterator and transform those errors into the desired format.

Let me know if I can provide more info

driv3r:
I assume we can get the path from the `Context.scope` I saw that it's available during validator compilation phase or should I look for something else?

Stranger6667:
Partially - it only changes when sub-folders are involved (via `$id` / `id` keywords), but it can be changed to keep track of all steps! :)

Also, it might be a bit tricky when the scope changes to another document (e.g. a remote one in the `$ref` keyword implementation), so we need to have all chunks that will lead to the root document.

But I think it is the right direction indeed :)

Stranger6667:
`flag` and `basic` styles were implemented via #274 

I'll put `detailed` & `verbose` to separate issues

PR #184, opened 2021-04-19
Add a `path` field to `ValidationError`

gavadinov:
### Summary
On validation error make the path to the problematic field available to the client. Currently the path is only a field in `ValidationError` but later it can be used to improve the error messages (related to #100, #144)
##### Schema
```json
{
   "type":"object",
   "properties":{
      "foo":{
         "type":"object",
         "properties":{
            "bar":{
               "type":"integer"
            }
         }
      }
   }
}
```
##### Input
```json
{
  "foo": {
    "bar": "some string"
  }
}
```

##### Produced Error
`path: [foo, bar]`

### DIFF
--- a/jsonschema/src/compilation/mod.rs
+++ b/jsonschema/src/compilation/mod.rs
@@ -4,6 +4,8 @@
 pub(crate) mod context;
 pub(crate) mod options;
 
+use std::{cell::RefCell, rc::Rc};
+
 use crate::{
     error::{CompilationError, ErrorIterator},
     keywords,
@@ -63,7 +65,14 @@ impl<'a> JSONSchema<'a> {
         let mut errors = self
             .validators
             .iter()
-            .flat_map(move |validator| validator.validate(self, instance))
+            .flat_map(move |validator| {
+                // The path capacity should be the average depth so we avoid extra allocations
+                validator.validate(
+                    self,
+                    instance,
+                    Rc::new(RefCell::new(Vec::with_capacity(6))).into(),
+                )
+            })
             .peekable();
         if errors.peek().is_none() {
             Ok(())
--- a/jsonschema/src/error.rs
+++ b/jsonschema/src/error.rs
@@ -51,6 +51,8 @@ pub struct ValidationError<'a> {
     pub instance: Cow<'a, Value>,
     /// Type of validation error
     pub kind: ValidationErrorKind,
+    /// Path of the property that failed validation
+    pub instance_path: Option<Vec<String>>,
 }
 
 /// An iterator over instances of `ValidationError` that represent validation error for the
@@ -177,28 +179,37 @@ pub enum TypeKind {
 impl<'a> ValidationError<'a> {
     pub(crate) fn into_owned(self) -> ValidationError<'static> {
         ValidationError {
+            instance_path: self.instance_path.clone(),
             instance: Cow::Owned(self.instance.into_owned()),
             kind: self.kind,
         }
     }
 
-    pub(crate) fn additional_items(instance: &'a Value, limit: usize) -> ValidationError<'a> {
+    pub(crate) fn additional_items(
+        instance_path: Vec<String>,
+        instance: &'a Value,
+        limit: usize,
+    ) -> ValidationError<'a> {
         ValidationError {
+            instance_path: Some(instance_path),
             instance: Cow::Borrowed(instance),
             kind: ValidationErrorKind::AdditionalItems { limit },
         }
     }
-    pub(crate) fn any_of(instance: &'a Value) -> ValidationError<'a> {
+    pub(crate) fn any_of(instance_path: Vec<String>, instance: &'a Value) -> ValidationError<'a> {
         ValidationError {
+            instance_path: Some(instance_path),
             instance: Cow::Borrowed(instance),
             kind: ValidationErrorKind::AnyOf,
         }
     }
     pub(crate) fn constant_array(
+        instance_path: Vec<String>,
         instance: &'a Value,
         expected_value: &[Value],
     ) -> ValidationError<'a> {
         ValidationError {
+            instance_path: Some(instance_path),
             instance: Cow::Borrowed(instance),
             kind: ValidationErrorKind::Constant {
                 expected_value: Value::Array(expected_value.to_vec()),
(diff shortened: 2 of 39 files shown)


Stranger6667:
Awesome @gavadinov !

I will take a look at it tomorrow :)

picklenerd:
I made a separate PR for the public error stuff.

https://github.com/Stranger6667/jsonschema-rs/pull/185

Stranger6667:
First of all, thank you for your efforts! I appreciate you working on this issue :)

I think that there should be an invariant - if any path stack is converted to JSON pointer, then looking up a value by this pointer should resolve into the validator's value. For example, let's take this schema:

```json
{
    "additionalItems": false,
    "items": [
        {
            "type": "string"
        }
    ]
}
```

If the additionalItems validator fails on `["a", 1]` , then the validation error should have `["additionalItems"]` as its path stack.

Python's `jsonschema` library behaves similarly:

```python
import jsonschema

try:
    jsonschema.validate(["a", 1], { "additionalItems": False , "items": [{"type": "string"}]})
except jsonschema.ValidationError as error:
    print(error.schema_path)

# Prints
# deque(['additionalItems'])
```

As far as I can see, it could be done via pushing needed keywords before (or just after) entering some validator compilation function.

**Tests**
Given the above invariant tests could be written like this:

```
#[test]
fn validator_path() {
    let validator_value = false;
    let schema = json!({ "additionalItems": validator_value , "items": [{"type": "string"}]});
    let value = json!(["a", 1]);
    let compiled = JSONSchema::compile(&schema).unwrap();
    let errors: Vec<ValidationError> = compiled.validate(&value).unwrap_err().collect();
    // Some routine to get a JSON pointer from the error schema path
    let pointer = format!("/{}", errors[0].path.as_ref().unwrap().join("/"));
    assert_eq!(schema.pointer(&pointer), Some(&json!(value)));
}
```
And then it could be parametrized with `test_case`. I'd appreciate some tests for common cases, including indexing within an array, so JSON pointers work. E.g., it will require `enumerate` in `ItemsArrayValidator::compile`

Here are some style nitpicks below

**Naming**
In the future, there could also be a path within the validated value. For this reason, I think that the path which is currently used in this PR could be named or derived from `schema_path`. In `CompilationContext` - `current_schema_path` and `schema_path` in all validation errors and validators.

**Arguments order**
I think that it will be nice to have this path information as the first argument whenever it is needed. E.g. it is the first argument to `ValidationError::additional_items` but the last one in compilation functions, e.g. in `AdditionalItemsBooleanValidator::compile`

gavadinov:
@Stranger6667 Thank you for the quick response. 

> if any path stack is converted to JSON pointer, then looking up a value by this pointer should resolve into the validator's value.

I'm not sure if this would fit the current implementation. The path that we generate now is not a path in the schema but a path in the JSON we are validating. Similar to the example here: #144 - this would let us know exactly which field in the JSON being validated failed validation.

For example the error produced by the example below would contain a path `[foo, bar]`:
##### Schema
```json
{
   "type":"object",
   "properties":{
      "foo":{
         "type":"object",
         "properties":{
            "bar":{
               "type":"integer"
            }
         }
      }
   }
}
```
##### Input
```json
{
  "foo": {
    "bar": "some string"
  }
}
```

We can add a new field that tracks the `schema_path` but that would have to touch more validators. Probably worth doing it separately?

Stranger6667:
> I'm not sure if this would fit the current implementation. The path that we generate now is not a path in the schema but a path in the JSON we are validating. Similar to the example here: #144 - this would let us know exactly which field in the JSON being validated failed validation.

Aha, I completely missed that! Then, the JSON pointer invariant should work for the validated instance :) 

At the moment it seems like intermediate arrays within the schema lead to incomplete paths.

Schema:

```json
{
    "type": "array",
    "items": [
        {
            "type": "string"
        },
        {
            "type": "object",
            "properties": {
                "foo": {
                    "type": "integer"
                }
            }
        }
    ]
}
```

Value: 

```json
["a", {"foo": "bar"}]
```
The path should be `["1", "foo"]` but it is `["foo"]` now.

Also on the naming then - `current_instance_path` and `instance_path` might fit better than to distinguish it with schema paths.

> We can add a new field that tracks the schema_path but that would have to touch more validators. Probably worth doing it separately?

Absolutely!

gavadinov:
Ahh good catch on the array indices :see_no_evil: I pushed a new revision with:
* Unit Tests - I added them to `error.rs` but let me know if there is a better place for them.
* Keeping track of the path through arrays.
* Renamed `path` to `instance_path`/`current_instance_path`.

Stranger6667:
Thanks!

However, I think that instance paths can't be unambiguously evaluated from schema only as the same validator can be used to validate different parts of the same input value. 

Schema:

```json
{
    "type": "array",
    "items": {
        "type": "string"
    }
}
```

Instance:

```json
["a", 42]
```

The current implementation returns `[]` as `instance_path`; however, it should be `["1"]`.

There is no information in the schema about how we can detect the particular failing index upfront. The `$ref` keyword also makes things more complicated, as any validator can be applied almost anywhere (including, e.g., using it from another arbitrary schema).

Therefore this information could only be obtained during validation and will have some runtime cost. But we can detect `schema_path` during the compilation phase.

gavadinov:
Ohh true. Is there a real use case for having the path be a valid JSON pointer? For my concrete use case I need to be able to obtain the name of the field that failed validation. So marking the arrays just with a `[]` would be fine: e.g.  `["foo", "bar", "[]", "baz"]`. Of course that is only for my use case, but I wonder if there is a real need for the concrete path that would justify the runtime cost :thinking:

Stranger6667:
The original reason is that Draft 2019-09 [defines](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-10.3.1) output formats this way, but I think that anything else than a complete JSON pointer won't be unambiguous in some cases. For example, it won't be possible to say which element in a 2-dimensional array is invalid.

Schema:
```json
{
    "type": "array",
    "items": {
        "type": "array",
        "items": {
            "type": "integer"
        }
    }
}
```

Instance:
```json
[
    [1, 2, 3],
    [4, "5", 6], 
    [7, 8, 9]
]
```

And other cases of nested schemas. In other words, if the error reporting aims to show the precise position inside the validated instance, it won't be possible.

gavadinov:
@Stranger6667 I reimplemented everything from scratch and now the path is tracked at runtime. The tests from the first iteration still pass because the public interface is the same. I also added test cases for your examples above and they pass now :tada:

Stranger6667:
This is amazing work! Thank you very much :) I will review it soon

Stranger6667 (approved):
I will merge it and then write some tests that will compare instance paths in errors from this library and Python's `jsonschema`. If they find something, then I'll fix them :) Then these changes will be released.

Thank you for your patience and efforts

gavadinov:
Perfect, thank you :partying_face:

Stranger6667:
The change is released in `0.7.0` :tada:

Stranger6667:
I forgot to implement a separate struct for `instance_path` to avoid some allocations. It is a breaking change and released in `0.8.0`. Sorry for the complications :(

This struct can be transformed to `Vec<String>` as it was before via the `.into_vec()` call:

```rust
error.instance_path.into_vec()
```

Or to JSON Pointer via `to_string()` or `Display` impl.

gavadinov:
That's amazing!! Thank you :tada:

PR #185, opened 2021-04-19
feat: Made error types public

picklenerd:
Allows users to handle and format their own errors. These changes are presented as an alternative to custom formatting.


Stranger6667 (approved):
Great! Thanks :)

Issue #183, opened 2021-04-08
patternProperties interferes with properties


duckontheweb:
It seems like the presence of `"patternProperties"` interferes with validation that should happen in `"properties"`. 

[This replit](https://replit.com/join/tottadhf-jonduckworth1) shows an example of an instance that should fail validation (and [does fail in jsonschemavalidator.net](https://www.jsonschemavalidator.net/s/zQTg6IT2)). If you comment out the `"patternProperties"` portion of the schema, it fails as expected.

I don't have a good enough understanding of the `jsonschema` internals, but it seems like this could be related to changes from #173 .

**UPDATE:** I updated the replit and the jsonschemavalidator link to provide a more minimal example.

duckontheweb:
I'm happy to take a crack at a PR, but I'm a complete Rust novice, so it might be a little ugly at first...

duckontheweb:
In the process of trying to debug this with different schema combinations I also uncovered a case where trying to compile a valid schema leads to a `SchemaError`.

Here is the schema:
```json
{
  "type": "object",
  "properties": {
    "eo:cloud_cover": {
      "type": "number"
    }
  },
  "patternProperties": {
    "^(?!eo:)": {}
  }
}
```
[Failing example here](https://replit.com/join/llvmdivv-jonduckworth1)

Stranger6667:
Hi @duckontheweb !

Thank you for an awesome bug report :) I'd be happy to review a PR, sure. Otherwise, I think that I'll have some time over the weekend to check why the bug is happening

duckontheweb:
The `CompilationError` happens because `regex` does not support look-around assertions (rust-lang/regex#127), which is not a bug in `jsonschema`. 

However, it seems like this is also the cause of the original issue. If I change the `"patternProperties"` regex to use the plain character set equivalent of that negative look-behind (`"^([^e]|e($|[^o])).*"`) then things work as expected ([example here](https://replit.com/join/sdalrkvh-jonduckworth1)).

It seems like the real bug is that compilation should have returned an error in the original example.

duckontheweb:
> The `CompilationError` happens because `regex` does not support look-around assertions ([rust-lang/regex#127](https://github.com/rust-lang/regex/issues/127)), which is not a bug in `jsonschema`.

[...]

PR #220, opened 2021-05-06
Add support for look-around patterns using fancy-regex

duckontheweb:
Uses [`fancy-regex`](https://docs.rs/fancy-regex/0.5.0/fancy_regex/) in place of [`regex`](https://docs.rs/regex/) for matching against all patterns in JSON schemas. 

`regex` is still used when [converting a RegEx for ECMA 262](https://github.com/duckontheweb/jsonschema-rs/blob/b689ac5b32c29db0a36cf7ad3890e78f0cea96e3/jsonschema/src/keywords/pattern.rs#L78) to take advantage of the `Regex::replace_all` method, which is not included in `fancy-regex`.  Since `fancy-regex` depends on `regex` anyway, this does not affect the final size of the compiled binary (thanks to [this commit](https://github.com/niklasfasching/check_timed_logs_fast/commit/fa2ec6ef77d3befd03d8674e0ceae24846b620a6) for the idea). 

I also removed the test from #214, since that pattern is supported in this PR.

### DIFF
--- a/jsonschema/src/keywords/pattern.rs
+++ b/jsonschema/src/keywords/pattern.rs
@@ -5,18 +5,16 @@ use crate::{
     paths::InstancePath,
     validator::Validate,
 };
-use regex::{Captures, Regex};
 use serde_json::{Map, Value};
 
-use std::ops::Index;
-
 lazy_static::lazy_static! {
-    static ref CONTROL_GROUPS_RE: Regex = Regex::new(r"\\c[A-Za-z]").expect("Is a valid regex");
+    // Use regex::Regex here to take advantage of replace_all method not available in fancy_regex::Regex
+    static ref CONTROL_GROUPS_RE: regex::Regex = regex::Regex::new(r"\\c[A-Za-z]").expect("Is a valid regex");
 }
 
 pub(crate) struct PatternValidator {
     original: String,
-    pattern: Regex,
+    pattern: fancy_regex::Regex,
 }
 
 impl PatternValidator {
@@ -43,12 +41,14 @@ impl Validate for PatternValidator {
         instance_path: &InstancePath,
     ) -> ErrorIterator<'a> {
         if let Value::String(item) = instance {
-            if !self.pattern.is_match(item) {
-                return error(ValidationError::pattern(
-                    instance_path.into(),
-                    instance,
-                    self.original.clone(),
-                ));
+            if let Ok(is_match) = self.pattern.is_match(item) {
+                if !is_match {
+                    return error(ValidationError::pattern(
+                        instance_path.into(),
+                        instance,
+                        self.original.clone(),
+                    ));
+                }
             }
         }
         no_error()
@@ -56,8 +56,10 @@ impl Validate for PatternValidator {
 
     fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
         if let Value::String(item) = instance {
-            if !self.pattern.is_match(item) {
-                return false;
+            if let Ok(is_match) = self.pattern.is_match(item) {
+                if !is_match {
+                    return false;
+                }
             }
         }
         true
@@ -66,12 +68,12 @@ impl Validate for PatternValidator {
 
 impl ToString for PatternValidator {
     fn to_string(&self) -> String {
-        format!("pattern: {}", self.pattern)
+        format!("pattern: {:?}", self.pattern)
     }
 }
 
 // ECMA 262 has differences
-fn convert_regex(pattern: &str) -> Result<Regex, regex::Error> {
+fn convert_regex(pattern: &str) -> Result<fancy_regex::Regex, fancy_regex::Error> {
     // replace control chars
     let new_pattern = CONTROL_GROUPS_RE.replace_all(pattern, replace_control_group);
     let mut out = String::with_capacity(new_pattern.len());
@@ -110,14 +112,16 @@ fn convert_regex(pattern: &str) -> Result<Regex, regex::Error> {
             out.push(current);
         }
     }
-    Regex::new(&out)
+    fancy_regex::Regex::new(&out)
 }
 
 #[allow(clippy::integer_arithmetic)]
-fn replace_control_group(captures: &Captures) -> String {
+fn replace_control_group(captures: &regex::Captures) -> String {
     // There will be no overflow, because the minimum value is 65 (char 'A')
     ((captures
-        .index(0)
+        .get(0)
+        .map(|m| m.as_str())
+        .expect("index 0 to return the whole match")
         .trim_start_matches(r"\c")
         .chars()
         .next()
@@ -147,7 +151,10 @@ mod tests {
     #[test_case(r"\\w", r"\w", true)]
     fn regex_matches(pattern: &str, text: &str, is_matching: bool) {
         let compiled = convert_regex(pattern).expect("A valid regex");
-        assert_eq!(compiled.is_match(text), is_matching);
+        assert_eq!(
+            compiled.is_match(text).expect("A valid pattern"),
+            is_matching
+        );
     }
 
     #[test_case(r"\")]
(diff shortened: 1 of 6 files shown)


jsonschema/src/keywords/pattern.rs
```diff
@@ -43,21 +41,25 @@ impl Validate for PatternValidator {
...
     }
 
     fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
         if let Value::String(item) = instance {
-            if !self.pattern.is_match(item) {
-                return false;
+            if let Ok(is_match) = self.pattern.is_match(item) {
+                if !is_match {
+                    return false;
+                }
```
Stranger6667, line 62: I think that it could be 

```suggestion
                return is_match
```

jsonschema/src/keywords/pattern.rs
```diff
@@ -43,21 +41,25 @@ impl Validate for PatternValidator {
...
             }
         }
         no_error()
     }
 
     fn is_valid(&self, _: &JSONSchema, instance: &Value) -> bool {
         if let Value::String(item) = instance {
-            if !self.pattern.is_match(item) {
-                return false;
+            if let Ok(is_match) = self.pattern.is_match(item) {
```
Stranger6667, line 59: What about the `Err` variant? I believe it should lead to `false`.

jsonschema/src/keywords/pattern.rs
```diff
@@ -43,21 +41,25 @@ impl Validate for PatternValidator {
         instance_path: &InstancePath,
     ) -> ErrorIterator<'a> {
         if let Value::String(item) = instance {
-            if !self.pattern.is_match(item) {
-                return error(ValidationError::pattern(
-                    instance_path.into(),
-                    instance,
-                    self.original.clone(),
-                ));
+            if let Ok(is_match) = self.pattern.is_match(item) {
```
Stranger6667, line 44: If `Err` is returned, we should indicate this backtracking problem to the user with a new `ValidationError` variant.

Stranger6667 (changes requested):
I like the direction and it looks good to me :)

I think that it is important to add a new variant to the `ValidationErrorKind` enum that will indicate the backtracking issue and use it in `PatternValidator::validate` and other cases when patterns are used (`additionalProperties`)

Stranger6667:
Maybe one extra thing - it will be nice to have a test case for `pattern` with backtracking

jsonschema/src/keywords/pattern.rs
duckontheweb: Yes, thanks!

jsonschema/src/keywords/pattern.rs
duckontheweb: Yes, good idea. I'll add this.

jsonschema/src/keywords/pattern.rs
duckontheweb: Yeah, I had a hard time reasoning out what would be expected here. I started typing out a response here and then came to same conclusion that it should lead to `false`. I'll update that.

duckontheweb:
@Stranger6667 How would you like me to handle the commits to address these comments. Should I squash them with the existing commit?

Stranger6667:
In the end, I plan to squash them into one, so feel free to push more :)

duckontheweb:
I think I addressed all of the suggestions above via the new commits. I also added matching and non-matching test cases to `pattern.rs`.

I had a thought that it would be good to add an integration test for the combination of `patternProperties`, `properties`, and `additionalProperties` that I mentioned in [this comment](https://github.com/Stranger6667/jsonschema-rs/issues/183#issuecomment-831473463), but I wasn't sure how to fit it in with the official JSON Schema test suite.

jsonschema/src/keywords/pattern.rs
```diff
@@ -43,35 +41,44 @@ impl Validate for PatternValidator {
...
+            return self.pattern.is_match(item).unwrap_or(false);
         }
         true
     }
 }
 
 impl ToString for PatternValidator {
     fn to_string(&self) -> String {
-        format!("pattern: {}", self.pattern)
+        format!("pattern: {:?}", self.pattern)
```
Stranger6667, line 76: Hm, is it required?

jsonschema/src/keywords/pattern.rs
```diff
@@ -110,14 +117,16 @@ fn convert_regex(pattern: &str) -> Result<Regex, regex::Error> {
...
 
 #[allow(clippy::integer_arithmetic)]
-fn replace_control_group(captures: &Captures) -> String {
+fn replace_control_group(captures: &regex::Captures) -> String {
     // There will be no overflow, because the minimum value is 65 (char 'A')
     ((captures
-        .index(0)
+        .get(0)
+        .map(|m| m.as_str())
+        .expect("index 0 to return the whole match")
```
Stranger6667, line 129: As the type didn't change, why are these changes needed?

jsonschema/src/keywords/pattern_properties.rs
```diff
@@ -80,7 +80,9 @@ impl ToString for PatternPropertiesValidator {
             "patternProperties: {{{}}}",
             self.patterns
                 .iter()
-                .map(|(key, validators)| { format!("{}: {}", key, format_validators(validators)) })
+                .map(|(key, validators)| {
+                    format!("{:?}: {}", key, format_validators(validators))
```
Stranger6667, line 84: Is the format change needed?

jsonschema/src/keywords/pattern_properties.rs
```diff
@@ -148,7 +150,7 @@ impl Validate for SingleValuePatternPropertiesValidator {
 impl ToString for SingleValuePatternPropertiesValidator {
     fn to_string(&self) -> String {
         format!(
-            "patternProperties: {{{}: {}}}",
+            "patternProperties: {{{:?}: {}}}",
```
Stranger6667, line 153: Is the format change needed?

jsonschema/src/keywords/pattern.rs
duckontheweb: These were required because `fancy_regex::Regex` implements `Debug` but not `Display`.

Stranger6667: Aha, got it :) then it obsoletes some of my other comments

jsonschema/src/keywords/pattern.rs
duckontheweb: Good catch. This got left over from my initial attempt to use `fancy_regex::Captures` here. Now that we're continuing to use `regex` this is no longer necessary, so I'll revert it.

jsonschema/src/keywords/pattern_properties.rs
duckontheweb: Same as above, this is required because `fancy_regex::Regex` implements `Debug` but not `Display`.

Stranger6667 (approved):

Stranger6667:
Looks great to me! :) Thank you for contributing. I appreciate it :)

I'll check it one more time tomorrow and probably will release these changes in the new release :)

duckontheweb:
I'm not sure what to make of that `clippy` failure. It doesn't show up when I run `cargo clippy --all-targets --all-features -- -D warnings` locally, but I'm on a Mac. I'll try running in an Ubuntu container as well to see if I can replicate it.

duckontheweb:
> Looks great to me! :) Thank you for contributing. I appreciate it :)
> 
> I'll check it one more time tomorrow and probably will release these changes in the new release :)

No problem, thanks for bearing with me on this. It's my first contribution to a Rust project and it's been really educational to see how this project is set up.

Stranger6667:
> I'm not sure what to make of that clippy failure. It doesn't show up when I run cargo clippy --all-targets --all-features -- -D warnings locally, but I'm on a Mac. I'll try running in an Ubuntu container as well to see if I can replicate it.

Hm, maybe it is connected to the recent Rust toolchain update :shrug: I'll apply a fix after merge :)

```
pub(crate) const fn backtrack_limit(
```

Stranger6667:
> No problem, thanks for bearing with me on this. It's my first contribution to a Rust project and it's been really educational to see how this project is set up.

I am glad to hear that! :)

Issue #199, opened 2021-04-29
Report schema paths in validation errors


Stranger6667:
When a validation error is reported, it contains only `instance_path` that points to the erroneous part of the input, but there is no information about what part of the schema caused this error.

The gist of the implementation is to keep a stack of strings while the input schema is compiled. The stack should be convertible to `paths::JSONPointer`.

The process for regular, non-composite validators (more on this below):
- When the compilation process enters a validator, push its relevant keyword name to the stack (e.g. `properties`);
- When the validator struct is created, clone this stack and store it within this struct;
- When the validator is compiled - pop from the stack
 
I think that pushing/popping could be implemented similarly to `instance_path` - a linked list without heap allocations. This way, the actual popping step is not needed at all.

Composite validators are slightly more complicated - they are single structs that implement the logic for multiple different keywords or their specific variants (e.g. [this one](https://github.com/Stranger6667/jsonschema-rs/blob/master/jsonschema/src/keywords/additional_properties.rs#L525) handles `patternProperties` and `additionalProperties: false`) which is done due to performance reasons.

To handle them, the stack should not be pushed during compilation, but rather during validation - if a validation error happens, the stored stack should be cloned, and the exact keyword variant should be pushed to it. And finally, it is moved to `ValidationError`.

Validators live in `keywords`.

A lot of work in this direction was done by @gavadinov in https://github.com/gavadinov/jsonschema-rs/commit/40149322bfdf5549a0150122a00fa66dfe770828, and I think that it could be a great base for implementing this feature.

@gavadinov, let me know if you are OK if somebody can base their work on your commit, or if you want to continue it yourself :)

DzikiChrzan:
May I take a swing on this?

Stranger6667:
Hi @DzikiChrzan!

Absolutely! Let me know if you have any questions on it :)

DzikiChrzan:
I've been studying the code and I can't fully understand the flow of the data.
First of all, how does `PathChunk` looks like and what `Index` in it is?
Do you recommend some end-to-end tests that I could run with gdb to actually see the flow? 

`Instance` is a path to json file, which is read via serde. At start there is called `validate` on Schema, which then calls `crate::validate(self, instance)` underneath. This is method from trait and I see that all keywords have those implementations, but I can't find out how validation by keyword is run on whole `instance_json`.

Stranger6667:
> First of all, how does `PathChunk` looks like and what Index in it is?

[...]

PR #225, opened 2021-05-09
Use meta-schemas to validate input schemas

### LINKED ISSUE #198: Use meta-schemas to validate input schemas
If the input schema is invalid during the schema compilation process, then `CompilationError` is emitted. It has no useful info inside, which leads to a bad UX in such cases.

It will be much better to use meta-schemas and existing `ValidationError` in such circumstances:

- Meta-schemas are known to be valid;
- Could be globally reused via `lazy_static`;
- Validation error will have much more verbose error messages;
- `instance_path` will be available in those errors;
- Adding `schema_path` to validation errors in the future will benefit the compilation process automatically.

Also, if meta-schemas will be sufficient, then `CompilationError` might be deleted. Even if not, then constructing `ValidationError` instead of them will provide much more useful output.

But it is a backward-incompatible change, which still seems worth it.

Roughly required steps are:
- Create different validators within `lazy_static!` in `compilation::options`. Probably it will require a slight refactoring there as meta-schemas are already included in the resulting binary - it will be nice to reuse those `serde_json::Value` values. It should be fine to use `expect` there as those validators are known to be valid;
- Make the input schema validation configurable - these schemas comply with themselves, which is a nice example of recursion - i.e. their validation depends on themselves :) So there should be a way to either force using some concrete meta-schema validation or disable it completely;
- Choose one of those validators within `compilation::options::CompilationOptions::compile` depending on the detected draft and the configuration logic from the previous step;
- Perform schema validation with this schema;
- Change the return type of the `compile` function to `ValidationError`;
- At this point, if schema validation is performed, it should not be possible to encounter some `CompilationError` down in compilation logic - but I am not 100% sure about it. Maybe we can create some "InvalidSchema" variant to `ValidationError` for such cases when the meta schema didn't catch the error. I assume that it is either pretty rare or even not possible, so having one descriptive extra variant in `ValidationError` seems like a good option instead of having `unreachable!` or `expect` that are panicking.

aedeny:
Closes #198.

### DIFF
--- a/jsonschema/src/compilation/options.rs
+++ b/jsonschema/src/compilation/options.rs
@@ -9,7 +9,7 @@ use crate::{
     schemas, ValidationError,
 };
 use ahash::AHashMap;
-use serde_json;
+
 use std::{borrow::Cow, fmt};
 
 lazy_static::lazy_static! {
@@ -36,7 +36,7 @@ lazy_static::lazy_static! {
 
     static ref META_SCHEMA_VALIDATORS: AHashMap<schemas::Draft, JSONSchema<'static>> = {
         let mut store = AHashMap::with_capacity(3);
-        const EXPECT_MESSAGE : &str = "Valid meta-schema!";
+        const EXPECT_MESSAGE: &str = "Valid meta-schema!";
         store.insert(
             schemas::Draft::Draft4,
             JSONSchema::compile(&DRAFT4).expect(EXPECT_MESSAGE)
--- a/jsonschema/src/compilation/options.rs
+++ b/jsonschema/src/compilation/options.rs
@@ -39,14 +39,15 @@ lazy_static::lazy_static! {
         const EXPECT_MESSAGE: &str = "Valid meta-schema!";
         store.insert(
             schemas::Draft::Draft4,
-            JSONSchema::compile(&DRAFT4).expect(EXPECT_MESSAGE)
+            JSONSchema::options().without_schema_validation().compile(&DRAFT4).expect(EXPECT_MESSAGE)
         );
         store.insert(
             schemas::Draft::Draft6,
-            JSONSchema::compile(&DRAFT6).expect(EXPECT_MESSAGE)
-        );store.insert(
+            JSONSchema::options().without_schema_validation().compile(&DRAFT6).expect(EXPECT_MESSAGE)
+        );
+        store.insert(
             schemas::Draft::Draft7,
-            JSONSchema::compile(&DRAFT7).expect(EXPECT_MESSAGE)
+            JSONSchema::options().without_schema_validation().compile(&DRAFT7).expect(EXPECT_MESSAGE)
         );
         store
     };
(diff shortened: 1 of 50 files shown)


jsonschema/src/compilation/options.rs
```diff
@@ -79,6 +102,11 @@ impl CompilationOptions {
         let resolver = Resolver::new(draft, &scope, schema, self.store.clone())?;
         let context = CompilationContext::new(scope, processed_config);
 
+        META_SCHEMA_VALIDATORS
+            .get(&draft)
+            .expect("existing draft!")
+            .validate(schema)
+            .expect("valid meta-schema!");
```
aedeny, line 109: @Stranger6667 Running `cargo check` outputs this:
```
PS C:\rust\jsonschema-rs\jsonschema> cargo check
    Checking jsonschema v0.9.0 (C:\rust\jsonschema-rs\jsonschema)
error[E0599]: the method `expect` exists for enum `Result<(), Box<dyn Iterator<Item = ValidationError<'_>> + Send + Sync>>`, but its trait bounds were not satisfied
   --> src\compilation\options.rs:109:14
    |
109 |               .expect("valid meta-schema!");
    |                ^^^^^^ method cannot be called on `Result<(), Box<dyn Iterator<Item = ValidationError<'_>> + Send + Sync>>` due to unsatisfied trait bounds
    | 
   ::: C:\Users\edeny\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\alloc\src\boxed.rs:165:1
    |
165 | / pub struct Box<
166 | |     T: ?Sized,
167 | |     #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global,
168 | | >(Unique<T>, A);
    | |________________- doesn't satisfy `_: Debug`
    |
    = note: the following trait bounds were not satisfied:
            `Box<dyn Iterator<Item = ValidationError<'_>> + Send + Sync>: Debug`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `jsonschema`

To learn more, run the command again with --verbose.
```

I will continue investigating it but I have a good feeling it's related to something small that I'm not familiar with in Rust yet. Please let me know if you have any suggestions regarding this.

Stranger6667: I think it could be something like this:

```rust
if let Some(mut errors) = META_SCHEMA_VALIDATORS
    .get(&draft)
    .expect("existing draft!")
    .validate(schema)
    .err()
{
    return Err(errors.next().expect("Should have at least one element"));
}
```

And it will require changing the return type a bit - `Result<JSONSchema<'a>, ValidationError<'a>>`. However, it leads to an error in `main.rs` that I don't know how to solve yet.

Stranger6667: In the `main.rs` file the problem could be solved like this:

```rust
fn validate_instances(instances: &[PathBuf], schema: PathBuf) -> BoxErrorResult<bool> {
    let mut success = true;

    let schema_json = fs::read_to_string(schema)?;
    let schema_json = serde_json::from_str(&schema_json)?;
    match JSONSchema::compile(&schema_json) {
        Ok(schema) => {
            for instance in instances {
                let instance_path_name = instance.to_str().unwrap();
                let instance_json = fs::read_to_string(&instance)?;
                let instance_json = serde_json::from_str(&instance_json)?;
                let validation = schema.validate(&instance_json);
                match validation {
                    Ok(_) => println!("{} - VALID", instance_path_name),
                    Err(errors) => {
                        success = false;

                        println!("{} - INVALID. Errors:", instance_path_name);
                        for (i, e) in errors.enumerate() {
                            println!("{}. {}", i + 1, e);
                        }
                    }
                }
            }
        }
        Err(error) => {
            println!("Schema is invalid. Error: {}", error);
            success = false
        }
    }
    Ok(success)
}
```

I.e. we match on the compilation result and in case of error print that the schema is not valid and return false to avoid lifetime issues with returning a reference to the value owned by that function. Otherwise, we run the old code in the `Ok` branch.

Stranger6667:
Hi @aedeny ! Let me know if I can help here :)

aedeny:
Hi @Stranger6667, I've just made a new commit but now I get lifetime issues after messing with lifetime specifiers in the `compile` methods. Not sure if I should have used `.clone()` instead, etc.

Other than this, please let me know if I'm on the right direction with the rest of the changes.

Stranger6667:
Yep, I think it is the right direction :) I'll dig deeper into this lifetime issue and let you know,
but generally, it seems like `context` and `value` / `parent` should have different lifetimes as `context` is created during the compilation process, and the reference to the input schema has a broader lifetime.

Stranger6667:
Hi @aedeny !

Sorry for the long silence, I finally managed to resolve that lifetime problem (from the third try), and I believe that the PR is ready for merging :) I also replaced `ComplationError` with `ValidationError`.

At some point in the future, it will be nice to replace the current usages of `ValidationError::schema` with the proper error kinds, but it will require #199 first.

Thank you so much for taking this! Let me know if I miss anything here

PR #226, opened 2021-05-10
Add schema paths to compilation context; this change requires changin…

DzikiChrzan:
Add schema paths to compilation context; this change requires changing mutability of compilation context

Work in progress

#199

### DIFF
--- a/jsonschema/src/keywords/additional_properties.rs
+++ b/jsonschema/src/keywords/additional_properties.rs
@@ -10,7 +10,7 @@ use crate::{
     compilation::{compile_validators, context::CompilationContext, JSONSchema},
     error::{error, no_error, ErrorIterator, ValidationError},
     keywords::{format_validators, CompilationResult, Validators},
-    paths::InstancePath,
+    paths::{InstancePath, JSONPointer},
     validator::Validate,
 };
 use ahash::AHashMap;
@@ -64,7 +64,7 @@ impl PropertiesValidatorsMap for BigValidatorsMap {
 }
 
 macro_rules! dynamic_map {
-    ($validator:tt, $properties:ident, $( $arg:tt ),* $(,)*) => {{
+    ($validator:tt, $properties:ident, $( $arg:expr ),* $(,)*) => {{
         if let Value::Object(map) = $properties {
             if map.len() < MAP_SIZE_THRESHOLD {
                 Some($validator::<SmallValidatorsMap>::compile(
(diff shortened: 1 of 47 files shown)


jsonschema/src/keywords/all_of.rs
```diff
@@ -14,10 +14,12 @@ pub(crate) struct AllOfValidator {
 
 impl AllOfValidator {
     #[inline]
-    pub(crate) fn compile(items: &[Value], context: &CompilationContext) -> CompilationResult {
+    pub(crate) fn compile(items: &[Value], context: &mut CompilationContext) -> CompilationResult {
         let mut schemas = Vec::with_capacity(items.len());
         for item in items {
+            context.schema_path.push(item.to_string());
```
Stranger6667, line 20: Here we need indexes within `items` - e.g., apply `enumerate` to the `items` above and store indexes like strings. Similar applies to `anyOf` validator implementation below.

jsonschema/src/keywords/additional_properties.rs
```diff
@@ -129,21 +129,25 @@ macro_rules! validate {
...
     Ok(properties)
 }
 fn compile_big_map(
     map: &Map<String, Value>,
-    context: &CompilationContext,
+    context: &mut CompilationContext,
 ) -> Result<BigValidatorsMap, CompilationError> {
     let mut properties = AHashMap::with_capacity(map.len());
     for (key, subschema) in map {
+        context.schema_path.push(subschema.to_string());
```
Stranger6667, line 148: It should be `key` instead of `subschema`

jsonschema/src/keywords/additional_properties.rs
```diff
@@ -129,21 +129,25 @@ macro_rules! validate {
 
 fn compile_small_map(
     map: &Map<String, Value>,
-    context: &CompilationContext,
+    context: &mut CompilationContext,
 ) -> Result<SmallValidatorsMap, CompilationError> {
     let mut properties = Vec::with_capacity(map.len());
     for (key, subschema) in map {
+        context.schema_path.push(subschema.to_string());
```
Stranger6667, line 136: It should be `key` instead of `subschema`

Stranger6667 (commented):
Yep, this is something I had in mind :) The invariant we should get in the end is - Looking up the resulting JSON Pointer in the input schema should lead to the sub-schema that failed validation.

For example:

```json
{
    "properties": {
        "foo": {
            "type": "string"
        }
    }
}
```

In the schema above, the stack for the `type` validator should look like this - `["properties", "foo"]`. Converting it to a JSON Pointer will give us `/properties/foo` and passing it to `schema.pointer` (assuming it is `serde_json::Value`) should return `{"type": "string"}`:

```rust
#[test]
fn test_schema_path() {
    let schema = json!({
        "properties": {
            "foo": {
                "type": "string"
            }
        }
    });
    let instance = json!({"foo": 42});
    let compiled = JSONSchema::compile(&schema).unwrap();
    let result = compiled.validate(&instance);
    let error = result.expect_err("Errors").next();
    let pointer = error.schema_path.to_string();
    assert_eq!(schema.pointer(&pointer), Some(&json!({"type": "string"})))
}
```

Stranger6667:
Hi @DzikiChrzan ! Let me know if I can help here :)

Stranger6667:
I made some progress on this (I hope you don't mind), and it seems like there is not that much left to do. My current todo items are:
- [ ] Write tests for each validator separately and fix cases that I missed (e.g., some combined validators might require some changes in `validate`)
- [ ] Compare `schema_path` with what Python's `jsonschema` is doing for the test corpus
- [ ] Extend Python's error messages with `schema_path` info
- [ ] Improve variable / method naming
- [ ] Introduce an `enum` for keywords to avoid some heap allocations during compilation & error reporting
- [ ] Minor optimizations to avoid extra work during compilation

Changing `ValidationError::schema` calls with the proper error types & paths will be outside of this PR.

PR #227, opened 2021-05-17
chore: Use Display formatting for Regex

duckontheweb:
Bumps `fancy-regex` to v0.6.0 (which now implements `Display` for the `Regex` struct) so we can go back to using `Display` formatting (see [this comment](https://github.com/Stranger6667/jsonschema-rs/pull/220#discussion_r627744671) from #220).


Stranger6667:
:tada:

Great! Thanks

Stranger6667 (approved):

Issue #246, opened 2021-07-10
Proposal: Support custom resolvers for external references


tamasfe:
Hi, I'm planning to use this library in [my project](https://github.com/tamasfe/taplo/issues/136), which is mostly used a constrained WASM environment.

This means that the filesystem, and http requests might not be directly available.

Is supporting a custom external resolver possible?

I propose the following:

Turn `Resolver` into a trait that could be used to resolve schemas in an unknown environment.

I'm thinking of something like this:

```rust
// The resolver that is used for external references.
//
// Internal references such as `#/definitions` and JSON pointers would still be handled internally.
trait Resolver<'resolver> {
    // Allow a custom error type with additional constraints if needed.
    type Error: std::error::Error;

    // The parent schema is provided so that the resolver can resolve relative refs
    // based on the schema `$id` or `id`.
    //
    // `&str` would be used instead of `Url` to support relative values like `./schema`.
    //
    // Returning a `Cow` would allow the resolver to cache schemas, however I'm not sure
    // whose responsibility caching should be.
    fn resolve(
        &self,
        schema: &Value,
        reference: &str,
    ) -> Result<Cow<'resolver, Value>, Self::Error>;
}
```

Then allow compilations with a custom resolver while providing a default one:

```rust
impl<'a> JSONSchema<'a> {
    pub fn compile_with(schema: &'a Value, resolver: impl Resolver<'a>) -> Result<JSONSchema<'a>, ValidationError<'a>> {
        // ...
    }

    pub fn compile(schema: &'a Value) -> Result<JSONSchema<'a>, ValidationError<'a>> {
        Self::compile_with(schema, DefaultResolver::new())
    }
}
```

Additionally an async version could be supported in a similar way which could help with resolution of many references. This would also be needed to support reqwest in WASM contexts.

<details>

```rust
trait AsyncResolver<'resolver> {
    type Error: std::error::Error;

    type ResolveFuture: Future<Output = Result<Cow<'resolver, Value>, Self::Error>> + 'resolver;

    fn resolve(
        &mut self,
        schema: &Value,
        reference: &str,
    ) -> Self::ResolveFuture;
}

impl<'a> JSONSchema<'a> {
    pub async fn compile_async_with(schema: &'a Value, resolver: impl AsyncResolver<'a>) -> Result<JSONSchema<'a>, ValidationError<'a>> {
        // ...
    }

[...]

Issue #212, opened 2021-05-03
Improve validators graph


Stranger6667:
The core idea behind this library is to store all keywords as a tree similar to the original schema JSON object. However, the implementation is not as efficient as I'd like it to be. Primarily, it has the following disadvantages:

- Each node's siblings are stored in a separate vector, which leads to a very fragmented graph that negatively affects cache locality;
- It doesn't allow cycles. The primary use case is the ` $ref` keyword, where some compilation logic is happening during validation. These things are cached, though, but the cost of `RWLock` is visible in flame graphs. Also, when logically the same validator is called recursively -> some nodes are re-created, but theoretically, they can be re-used.

This issue is the starting point to rethink the implementation and maybe to implement something better.

Here are some thoughts I have in mind:
- It might be possible to use an arena to allocate all boxed validators, then navigating could be done with indexes (maybe);
- An interesting [article](http://smallcultfollowing.com/babysteps/blog/2015/04/06/modeling-graphs-in-rust-using-vector-indices/) on using indexes in graphs;
- It might be nice to have some tool to display the validators graph.
- More [ideas](https://github.com/nrc/r4cppp/blob/master/graphs/README.md)

Later I'll draw the current implementation, so all the overhead is more visible.

Stranger6667:
At the moment, I think that the following layout will be a good step in this direction:

- Store validators in a vector
- Store metadata (relative_path / absolute_path) in a separate vector
- Pass `&[BoxedValidator]` to each validator in `is_valid` / `validate` / `apply` so each validator can call dependent validators
- Use indexes to access dependent validators/metadata (as a single one or a range)

This should improve cache locality as all validators will be in the same vector + remove costs for following references between the validation nodes (it will be array access by index) + `is_valid` won't have extra costs for loading larger `SchemaNode` as it will be just boxed validator

PR #247, opened 2021-07-12
WIP: Validator Layout and Resolver

tamasfe:
Hi @Stranger6667, I've done some experiments I'd like to share.

This PR touches on the topics of #246 and #212. This isn't intended to be merged as is (or at all), I'm just sharing my thoughts.

#### Validator graph

I made a validator layout based on the arena+index/reference approach to support cycles, for this I made two opaque types:

[`ValidatorBuf`](https://github.com/tamasfe/jsonschema-rs/blob/476404acb61d7f029790eb4733f5431eca70caa4/jsonschema/src/compilation/context.rs#L105) and [`ValidatorRef`](https://github.com/tamasfe/jsonschema-rs/blob/476404acb61d7f029790eb4733f5431eca70caa4/jsonschema/src/compilation/context.rs#L120), the idea is that we collect all `ValidatorBuf`s first in the `CompilationContext` then finally store them in `JSONSchema`, and the validator tree itself only contains references.

In order to be able to do this, the only way to obtain a `ValidatorRef` during compilation is through the `CompilationContext` [here](https://github.com/tamasfe/jsonschema-rs/blob/476404acb61d7f029790eb4733f5431eca70caa4/jsonschema/src/compilation/context.rs#L89). Currently the only thing it does is adding the validator to the pool, and returning a reference that can can be used in the tree.

This method also enables further operations/optimizations of validators during compilation since we have all of them at one place no matter how deep the tree is.

The reason why `ValidatorBuf` and `ValidatorRef` are wrapper types is that we can freely change the underlying implementation, currently I used `Arc` for storing the validators and `Weak` inside `ValidatorRef` in debug mode to make sure that everything works,  then finally `Box` and raw pointers in release mode. I'm confident about the safety of the latter as we always keep all validators in `JSONSchema`, and everything currently is synchronous.

#### Resolver

I started implementing an external resolver, while turning the current `Resolver` to an `InternalResolver`, that will solely resolve JSON pointers (and fragments). I've figured that no matter what we will have to carry it around in the context due to the architecture, and I still have no idea whether it would be better to just box it or carry the generic type around everywhere.

I also haven't figured out how to support both sync and async compilations as it will involve a lot of boilerplate either way, the best idea I have so far is just a macro that wraps compile functions that can recurse into `compile_validators` to be both sync and async.

#### Other thoughts

It would be nice to split the benchmarks up into parts, currently it takes ages especially because it contains benchmarks not related to `jsonschema`, I had unfortunately no time to run them and do this refactor, so currently I can only guess the performance implications of the changes I made.

I've experimented with reusing validators, meaning that if more validators would do the same thing, the references in the tree would simply point to the same instance, making the pool more compact and cache-friendly. This is currently not possible because the validators also contain their path inside the schema.

_edit: I've also changed all ToString implementations to Display, when I was halfway into it I understood why it hasn't been done, but decided to finish it it anyway ..._


tamasfe:
Another important change I forgot to mention is `#[non_exhaustive]` on the errors, this is a breaking API change now, however new validators and their error types can be added in any release, so it would prevent further breaking changes.

Stranger6667:
@tamasfe 

Amazing work! Thank you for sharing it! :)

> edit: I've also changed all ToString implementations to Display, when I was halfway into it I understood why it hasn't been done, but decided to finish it it anyway ...

I cherry-picked some changes there to reduce the size of the diff :) Feel free to rebase (the changes are almost identical, so you can choose your branch to resolve conflicts). Hopefully, there is not much trouble.

> Validator graph

I like the idea and feel like it is a step in the right direction.

To elaborate on the changes, I'd like to share my understanding of the status quo and the proposed changes (I'll omit the trait bounds for brevity). The current implementation looks like this (please, correct me if I am wrong):

JSONSchema-validators-current

1. `JSONSchema.validators` is a vector of `Box<dyn Validate>`, i.e. a pointer to heap allocation of other pointers (B_11, B_12, ...)
2. Each box there leads to a validator that may have `Vec<Box<dyn Validate>>` with similar pointers to other heap allocations (B_21, B_22, ...)

So:
- a separate vector for each level;
- Each `Box<dyn Validate>` goes to a different allocation. 

Let's take this schema - `{"items": {"type": "string"}}`, so the way references are followed is roughly like this:
1. Follow the reference to the slice of `Box<dyn Validate>` [1] (iterate over `self.validators`)
2. Deref `Box` that wraps `dyn Validate` leading to `ItemsObjectValidator` [2]
3. Follow `ItemsObjectValidator.validators` to another slice of `Box<dyn Validate>` [3]
4. Deref `Box` and get `StringTypeValidator` [4]

\+ dynamic dispatch costs

With the proposed changes:
1. `JSONSchema.validators` is effectively a vector of `*const (dyn Validate)`, which seems the same number of indirections as before (a pointer to a slice of pointers to trait objects).
2. On this level, validators may have a vector of pointers that lead to the same allocation of trait objects.

So, there are still vectors inside the concrete validator types, but each box will lead to the same slice of trait objects. I.e. the example above will have 1 allocation less ([1] and [3] will lead to the same slice).

If my understanding is correct then, I think it is a really great step forward :) Please, let me know if I am missing anything there.

Btw, I was thinking whether it could be even more compact like this:

JSONSchema-validators-maybe

I.e. if each concrete validator type will hold `&[&dyn Validate]` (or `&[Box<dyn Validate>]` or `&[* const dyn Validate]`) that will lead to the same allocation. But it is probably could be discussed later.

> Resolver

> I started implementing an external resolver, while turning the current Resolver to an InternalResolver, that will solely resolve JSON pointers (and fragments). I've figured that no matter what we will have to carry it around in the context due to the architecture, and I still have no idea whether it would be better to just box it or carry the generic type around everywhere.

My first instinct is to have a generic there, but I need to check the code in more detail to have a better feeling on this :)

> I also haven't figured out how to support both sync and async compilations as it will involve a lot of boilerplate either way, the best idea I have so far is just a macro that wraps compile functions that can recurse into compile_validators to be both sync and async.

Sounds good to me

> It would be nice to split the benchmarks up into parts, currently it takes ages especially because it contains benchmarks not related to jsonschema, I had unfortunately no time to run them and do this refactor, so currently I can only guess the performance implications of the changes I made.

I'll work on splitting the benchmarks (I saw [a nice example](https://github.com/PyO3/pyo3/tree/main/benches) of that in PyO3 recently), so it is easier to assess the changes.

> I've experimented with reusing validators, meaning that if more validators would do the same thing, the references in the tree would simply point to the same instance, making the pool more compact and cache-friendly. This is currently not possible because the validators also contain their path inside the schema.

Interesting idea! But yeah, the presence of paths limits this option.

> Another important change I forgot to mention is #[non_exhaustive] on the errors, this is a breaking API change now, however new validators and their error types can be added in any release, so it would prevent further breaking changes.

Cool! I didn't think about it. I believe it could be factored out and merged separately.

Generally, I think it will worth it to split the changes into separate PRs starting with breaking the cycles & resolving everything during the compilation phase, then go to resolvers.

tamasfe:
> If my understanding is correct then, I think it is a really great step forward :)

Thanks, yes. This is the gist of it.

The main goal is that if we encounter validators for schemas that already exist (e.g. ones with $id or resolved through $ref) in `compile_validators`, we can simply return references to them instead of calling compile, so a validator can even contain itself as a child. (this logic is not in this PR)

The main issue I've encountered with this is the handling of paths. What happens to paths after switching schema roots via $refs? Are they indefinitely appended, or is the root reset?

If the root is always calculated from the schema with $id, then this isn't an issue, but if we want to keep the entire path starting from the first schema, this can get tricky. Although I don't recommend the latter as paths could get extremely long for recursive validations.

Either way I would consider moving the schema path from the validator into the tree somehow, so validators would look something like `Vec<(SchemaPath, ValidatorRef)>`. Currently validators cannot get reused because they contain the schema path, so the array gets blown up with validators like `type: integer`, which is pretty much just the same function.

tamasfe:
> Generally, I think it will worth it to split the changes into separate PRs starting with breaking the cycles & resolving everything during the compilation phase, then go to resolvers.

Yes, the plan was creating separate PR-s that are properly tested and documented, but I don't have time for it right now, so I decided to submit this as is for discussions. Feel free to pick changes you like and continue what I started and create the said PR-s, as I have very limited time during weekdays.

Stranger6667:
@tamasfe 
I am so sorry, I completely missed your previous message :(

> The main goal is that if we encounter validators for schemas that already exist (e.g. ones with $id or resolved through $ref) in compile_validators, we can simply return references to them instead of calling compile, so a validator can even contain itself as a child. (this logic is not in this PR)

That's a really cool idea! :)

> The main issue I've encountered with this is the handling of paths. What happens to paths after switching schema roots via $refs? Are they indefinitely appended, or is the root reset?

In the current implementation, paths are joined together, so recursion is reflected in resulting errors (not in validator structs itself). So, the root is reset and is local in new validators, but in the end, all chunks are joined together.

It is close to the [Keyword Relative Location](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.12.3.1) definition from Draft 2020-12 but does not include `$ref` in the resulting paths. Here I followed the way it is implemented in Python's [jsonschema](https://github.com/Julian/jsonschema) library. 

> If the root is always calculated from the schema with $id, then this isn't an issue, but if we want to keep the entire path starting from the first schema, this can get tricky. Although I don't recommend the latter as paths could get extremely long for recursive validations.

Agree, probably to have proper relative locations we still need some validation-time path building.

> Either way I would consider moving the schema path from the validator into the tree somehow, so validators would look something like Vec<(SchemaPath, ValidatorRef)>. Currently validators cannot get reused because they contain the schema path, so the array gets blown up with validators like type: integer, which is pretty much just the same function.

Yep, good idea!

tamasfe:
@Stranger6667 hey, no problem :)

Unfortunately I won't have time to continue this in the foreseeable future as even my weekends are chaotic right now. I'll eventually make some time to pitch in though!

I think the next thing I'd do is factoring out schema paths from validators into the tree so that validators are really just context-independent reusable functions and only then work on the arena-like allocations and async boilerplates.

Stranger6667:
I will close this one, as I work on a rewrite where some of these ideas are incorporated, so in the end, all refs are resolved during the compilation phase. The graph itself is in the [CSR format](https://en.wikipedia.org/wiki/Sparse_matrix), which should be very cache-friendly & should be traversed fast enough. It is not finished but I plan to share a WIP PR here in a week or so.

PR #273, opened 2021-08-11
added uuid lib for format validation

rahulahoop:


jsonschema/src/keywords/format.rs
```diff
@@ -401,7 +414,8 @@ pub(crate) fn compile<'a>(
             "idn-email" => Some(IDNEmailValidator::compile(context)),
             "idn-hostname" if draft_version == Draft::Draft7 => {
                 Some(IDNHostnameValidator::compile(context))
-            }
+            },
+            "uuid" => Some(UUIDValidator::compile(context)),
```
Stranger6667, line 418: Please, put it behind `#[cfg(feature = "draft201909")]` macro & a condition as some other Draft 2019-09 validators in this match

Stranger6667 (changes requested):
Great! Thanks! 

You need to adjust the commit message a bit :) We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

Stranger6667:
Additionally, please, fix `rustfmt` & `clippy` failures

jsonschema/src/keywords/format.rs
Stranger6667: It might look like this:

```rust
#[cfg(feature = "draft201909")]
"uuid" if draft_version == Draft::Draft201909 => {
    Some(UUIDValidator::compile(context))
}
```

rahulahoop: yes but then my test fails because in test context there is not draft_version configuration

Stranger6667: You need to run tests with `--all-features` or `--features draft201909`

rahulahoop: ive updated my tests to use the tests_util file. im not sure the implementation is working correctly

```
        let passing_instance = json!("f308a72c-fa84-11eb-9a03-0242ac130003");
        let failing_instance = json!("º∆ª©");

        tests_util::is_valid(&schema, &passing_instance);
        tests_util::is_not_valid(&schema, &failing_instance);
```

"is not valid" makes the test fail, because i expect uuid::fromStr to return an error (and it does) but i think compile isnt calling my implementation correctly ?

rahulahoop: basically no matter what string i feed it, is_valid is always true

rahulahoop: yea when i debug, its not calling the UUIDValidator method i implemented :(

Stranger6667: Oh, I forgot to mention, but you need to use `JSONSchema::options().with_draft(Draft::Draft201909)` and then validate input, as this format is only available under this draft

rahulahoop: aha interesting.. so the test will need to change once/if the draft is removed?

i hit the debug point tho !

rahulahoop:
ive made the changes we discussed. ran clippy and fmt locally and seem to be fine!
Thanks for your help, this was really interesting !!

Stranger6667 (approved):
Awesome! Thanks for your contribution :)

PR #274, opened 2021-08-13
Output formats

alexjg:
The 2020-12 spec specifies a few things that are relevant for this PR.

- A "basic" output format which specifies how errors and annotations from failing or successful validations respectively should be structured
- A detailed specification of what annotations should be produced by "applicator" keywords - which are effectively control flow operators such as `oneOf`
- Explicitly specifying that unknown keywords must be retained and made available as annotations. For example the schema `{"name": "string","character": "narrator"}` must produce the annotation `{"character":"annotation"}`

This has implications which have led to the changes in this PR. Most importantly, we must collect unmatched keywords and store them in the validator tree so we can later produce annotations based on them. This has led me to introduce the `SchemaNode` struct, which replaces the `type Validators = Vec<Box<dyn Validator>>` type alias we were using to represent nodes in the schema tree previously.

Secondly, the `validate` method only returns errors. Rather than modifying that API and introducing severe breakage for existing clients I have introduced another method on the `Validate` trait called `apply`, which returns a data structure which can be turned into a `BasicOutput`, which in turn implements `serde::Serialize` in a manner conforming to the basic output format of the spec.

## `SchemaNode`

`SchemaNode` can be found in `src/schema_node.rs` and looks like this:

```rust
pub(crate) struct SchemaNode {
    validators: NodeValidators,
    relative_path: JSONPointer,
    absolute_path: Option<AbsolutePath>,
}
```

The `NodeValidators` enum captures the different contexts in which we create a node in the validator tree, see the docs for more information on that. In particular `NodeValidators` is where we store any unmatched keywords. We also record the relative and possible absolute path of the `Node` for use in producing the output formats.

We are able to implement `Validate` for `SchemaNode`.

## `Validate::apply`

We add the `apply` method which returns an output format including any annotations the validator may want to add. This may at first seem like it would require us to implement `apply` for _all_ of the existing keywords. However, this is unnecessary due to the introduction of the `SchemaNode` struct. Any applicator keywords have references to `SchemaNode`s containing their children, rather than directly to the child `Validate`s, this means that we can add a default implementation of `apply` in terms of `validate` and the `apply` implementation on `SchemaNode` takes care of adding annotations from unmatched keywords. This means that we must only implement `apply` for applicator keywords, which reduces the work somewhat.

Currently the implementation of `apply` for most keywords is very similar to the implementation of `validate`. It would be possible to reimplement `validate` in terms of `apply` for applicator keywords to reduce this duplication but I think that this is likely to have serious performance implications. To achieve good performance here I think we may need to look at using an arena allocator like `bumpalo` for collecting annotations and errors.

Note that there are extensive tests for the output formats in `tests/output.rs`

# Performance

Unfortunately introducing the `SchemaNode` does introduce some overhead. For `is_valid` this overhead is very low. Most benchmarks increased by less than 10 nanoseconds. I investigated this in detail and the generated assembly does include quite a few more instructions just for dereferencing the the additional data structures involved. I've managed to remove some of this for cases where a `SchemaNode` only has one child - which is the common case. I think increases of less than 10ns are acceptable.

`validate` unfortunately is significantly more affected. This is due to additional allocations which are necessary to make the lifetimes of `validate` match up when returning from `SchemaNode` - which wasn't a problem when were directly iterating over `Validators` before. As with the code duplication problem I think that the way to solve this is probably to investigate using an arena allocator.

I've attached before and after criterion baselines, but the top level performance changes are as follows:

| Benchmark Category | Fractional Change |
--|--|
| `validate` | 66% |
| `compile`  | 8%  |
| `is_valid` | 14% |

Note that if we exclude benchmarks which changed by less than 10ns we get

| Benchmark Category | Fractional Change |
--|--|
| `validate` | 66% |
| `compile`  | 8%  |
| `is_valid` | 9%  |

With several `is_valid` benchmarks actually decreasing (most of the format related ones). In fact, the only `is_valid` benchmarks which increased by more than 10ns are:

| Benchmark                      | Change % | Increased by |
---|---|---|
| jsonschema-rs CITM is_valid    | 4%       | 42µs         |
| jsonschema-rs geojson is_valid | 3%       | 52µs         |
| jsonschema-rs openapi is_valid | 9%       | 700µs        |
| jsonschema-rs swagger is_valid | 10%      | 700µs        |

I've looked into these a little and timings seem dominated by drop implementations. Regardless, these seem like acceptable numbers to me in return for supporting the basic output format. I also think performance can be improved back to the original number but I don't want to introduce an even larger change than this already large PR.

[baselines.zip](https://github.com/Stranger6667/jsonschema-rs/files/6983914/baselines.zip)

### DIFF
--- a/jsonschema/src/compilation/context.rs
+++ b/jsonschema/src/compilation/context.rs
@@ -1,5 +1,6 @@
 use super::options::CompilationOptions;
 use crate::{
+    compilation::DEFAULT_SCOPE,
     paths::{InstancePath, JSONPointer, PathChunk},
     schemas,
 };
@@ -7,18 +8,74 @@ use serde_json::Value;
 use std::borrow::Cow;
 use url::{ParseError, Url};
 
+static DEFAULT_SCHEME: &str = "json-schema";
+
 /// Context holds information about used draft and current scope.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct CompilationContext<'a> {
-    pub(crate) scope: Cow<'a, Url>,
+    base_uri: BaseUri<'a>,
     pub(crate) config: &'a CompilationOptions,
     pub(crate) schema_path: InstancePath<'a>,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) enum BaseUri<'a> {
+    /// A base URL was specified, either because we know a reasonable base URI where we retrieved the
+    /// schema or because a URI was specified via $id
+    Known(Cow<'a, Url>),
+    /// We don't know a base URI for this schema
+    Unknown,
+}
+
+impl<'a> BaseUri<'a> {
+    pub(crate) fn with_new_scope(&self, new_scope: &str) -> Result<Self, ParseError> {
+        let options = match self {
+            BaseUri::Known(u) => Url::options().base_url(Some(u)),
+            BaseUri::Unknown => Url::options().base_url(Some(&DEFAULT_SCOPE)),
+        };
+        Ok(options.parse(new_scope)?.into())
+    }
+}
+
+impl<'a> From<Option<Url>> for BaseUri<'a> {
+    fn from(u: Option<Url>) -> Self {
+        u.map_or(BaseUri::Unknown, |u| u.into())
+    }
+}
+
+impl<'a> From<&'a Url> for BaseUri<'a> {
+    fn from(u: &'a Url) -> Self {
+        if u.scheme() == DEFAULT_SCHEME {
+            BaseUri::Unknown
+        } else {
+            BaseUri::Known(Cow::Borrowed(u))
+        }
+    }
+}
+
+impl<'a> From<Url> for BaseUri<'a> {
+    fn from(u: Url) -> Self {
+        if u.scheme() == DEFAULT_SCHEME {
+            BaseUri::Unknown
+        } else {
+            BaseUri::Known(Cow::Owned(u))
+        }
+    }
+}
+
+impl<'a> From<&'a BaseUri<'a>> for Cow<'a, Url> {
+    fn from(uri: &BaseUri<'a>) -> Self {
+        match uri {
+            BaseUri::Unknown => Cow::Borrowed(&DEFAULT_SCOPE),
+            BaseUri::Known(u) => u.clone(),
+        }
+    }
+}
+
 impl<'a> CompilationContext<'a> {
-    pub(crate) const fn new(scope: Url, config: &'a CompilationOptions) -> Self {
+    pub(crate) const fn new(scope: BaseUri<'a>, config: &'a CompilationOptions) -> Self {
         CompilationContext {
-            scope: Cow::Owned(scope),
+            base_uri: scope,
             config,
             schema_path: InstancePath::new(),
         }
@@ -37,15 +94,14 @@ impl<'a> CompilationContext<'a> {
     #[inline]
     pub(crate) fn push(&'a self, schema: &Value) -> Result<Self, ParseError> {
         if let Some(id) = schemas::id_of(self.config.draft(), schema) {
-            let scope = Url::options().base_url(Some(&self.scope)).parse(id)?;
             Ok(CompilationContext {
-                scope: Cow::Owned(scope),
+                base_uri: self.base_uri.with_new_scope(id)?,
                 config: self.config,
                 schema_path: self.schema_path.clone(),
             })
         } else {
             Ok(CompilationContext {
-                scope: Cow::Borrowed(self.scope.as_ref()),
+                base_uri: self.base_uri.clone(),
                 config: self.config,
                 schema_path: self.schema_path.clone(),
             })
@@ -56,7 +112,7 @@ impl<'a> CompilationContext<'a> {
     pub(crate) fn with_path(&'a self, chunk: impl Into<PathChunk>) -> Self {
         let schema_path = self.schema_path.push(chunk);
         CompilationContext {
-            scope: Cow::Borrowed(self.scope.as_ref()),
+            base_uri: self.base_uri.clone(),
             config: self.config,
             schema_path,
         }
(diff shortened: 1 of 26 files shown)


jsonschema/src/output.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+impl<'a> serde::Serialize for AnnotationsInner<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        match self {
+            Self::UnmatchedKeywords(kvs) => {
+                let stdmap: std::collections::HashMap<String, serde_json::Value> =
```
Stranger6667, line 241: Would it work with `kvs` directly if we enable the `serde` feature from `ahash`?

jsonschema/src/paths.rs
```diff
@@ -41,11 +41,21 @@ impl JSONPointer {
...
         &self.0
     }
 }
 
+impl serde::Serialize for JSONPointer {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.to_string().as_str())
```
Stranger6667, line 55: I am wondering whether it would be possible to avoid allocating `String` here and write directly to the underlying receiver & whether it worth the complexity

jsonschema/src/validator.rs
```diff
@@ -15,6 +34,127 @@ pub(crate) trait Validate: Send + Sync + core::fmt::Display {
...
+    ///     // Here we use the `From<BasicOutput> for PartialApplication impl
+    ///     if_result.into()
+    /// } else {
+    ///     self.else_schema
+    ///         .apply_rooted(schema, instance, instance_path)
+    ///         .into()
+    /// }
+    /// ```
+    ///
+    /// `BasicOutput` also implments `Sum<BasicOutput>` and `FromIterator<BasicOutput<'a>> for PartialApplication<'a>`
```
Stranger6667, line 79:
```suggestion
    /// `BasicOutput` also implements `Sum<BasicOutput>` and `FromIterator<BasicOutput<'a>> for PartialApplication<'a>`
```

jsonschema/src/validator.rs
```diff
@@ -15,6 +34,127 @@ pub(crate) trait Validate: Send + Sync + core::fmt::Display {
...
+        if errors.is_empty() {
+            PartialApplication::valid_empty()
+        } else {
+            PartialApplication::invalid_empty(errors)
+        }
+    }
+}
+
+/// The result of applying a validator to an instance. As explained in the documentation for
+/// `Validator::apply` this is a "partial" result because it does not include information about
```
Stranger6667, line 100: Should it be `Validate::apply` instead?

jsonschema/src/schema_node.rs
```diff
@@ -0,0 +1,429 @@
...
+    JSONSchema,
+};
+use ahash::AHashMap;
+use std::fmt;
+
+/// A node in the schema tree, returned by [`compile_validators`]
+#[derive(Debug)]
+pub(crate) struct SchemaNode {
+    validators: NodeValidators,
+    relative_path: JSONPointer,
```
Stranger6667, line 17: In many validators, there are `schema_path` fields (e.g. `OneOfValidator`). Do we need them when `SchemaNode` is also present? Mostly thinking in terms of whether the path we need for `ValidationError` could be taken from here

jsonschema/src/schema_node.rs
```diff
@@ -0,0 +1,429 @@
...
+                    child_results.insert(0, self.annotation_at(instance_path, annotations));
+                };
+                BasicOutput::Valid(child_results)
+            }
+            PartialApplication::Invalid {
+                errors,
+                mut child_results,
+            } => {
+                for error in errors {
+                    child_results.insert(0, self.error_at(instance_path, error));
```
Stranger6667, line 138: Would it make sense to use `VecDeque` here instead?  to avoid moving elements

jsonschema/src/schema_node.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    fn apply<'a>(
+        &'a self,
+        schema: &JSONSchema,
+        instance: &serde_json::Value,
+        instance_path: &InstancePath,
+    ) -> PartialApplication<'a> {
+        //let mut success_results: Vec<OutputUnit<Annotations>> = Vec::new();
+        //let mut annotations = None;
```
Stranger6667, line 331: Seems like a leftover

Stranger6667 (commented):
That looks amazing to me! Thank you so much! Impressive and well-thought changes :)

I left a couple of minor comments

Stranger6667:
Also, thank you for a great write-up and docs! :) It helped me a lot during reviewing

Stranger6667:
On the performance side of things. I think that using an arena allocator is the way to go, indeed, but those performance changes are acceptable :) There are some more ideas & discussions on this matter in #212 and #247

jsonschema/src/paths.rs
alexjg: Today I learned about `Serializer::collect`, which does exactly this https://docs.serde.rs/serde/ser/trait.Serializer.html#method.collect_str

jsonschema/src/compilation/context.rs
```diff
@@ -1,24 +1,82 @@
...
 use url::{ParseError, Url};
 
+static DEFAULT_SCHEME: &str = "json-schema";
+
 /// Context holds information about used draft and current scope.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct CompilationContext<'a> {
-    pub(crate) scope: Cow<'a, Url>,
+    //scope: Cow<'a, Url>,
```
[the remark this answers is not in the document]
alexjg: Fixed

jsonschema/src/output.rs
alexjg: It does indeed!

jsonschema/src/validator.rs
alexjg: Fixed

jsonschema/src/validator.rs
alexjg: Yup, fixed.

jsonschema/src/schema_node.rs
alexjg: This definitely could be done. My instinct is that the best way to actually do it is to start trying to reimplement `validate` using `apply` for applicator keywords. That reduces some of the other logic duplication in applicator keywords and provides a nice avenue to start experimenting with an arena allocator for error collection. i.e I think we should leave it out of this PR and address it shortly in another PR.

I'm happy to address in this PR if you feel it shouldn't be left until later though.

jsonschema/src/schema_node.rs
alexjg: Good shout, fixed.

jsonschema/src/schema_node.rs
alexjg: Fixed

alexjg:
@Stranger6667 thanks for the review! I think I've addressed most of your points. I've also rebased into a single conventional commit so this is ready to merge at your convenience :slightly_smiling_face:

jsonschema/src/schema_node.rs
Stranger6667: Cool! I think it could be done separately :)

Issue #215, opened 2021-05-05
Use the original value in error formatting for some numeric validators


Stranger6667:
Validating `3.0` against `{"exclusiveMaximum": 3.0}` will give `3.0 is greater than or equal to the maximum of 3`. But the last number in the error message should be `3.0`. It affects `exclusiveMaximum`, `exclusiveMinimum`, `minimum`, `maximum`.

djmitche:
I'll give this a shot, if that's OK :)

Stranger6667:
@djmitche 
Absolutely :) Thank you

Let me know if I can help

djmitche:
It looks like this would be passing a different value for `limit` in `src/keywords/exclusive_maximum.rs` (and similarly for the other limits).  Two questions:

 * For f64, this will mean printing a parsed value that may not match the source document (e.g., `3.0000` -> `3.0`).  I think that's the best we can do, given a `json::Value` in the schema.  Is that OK?
 * Presently, the `limit` parameter is an f64, I think as a way of encompassing all possible values.  Would it make most sense to switch this to a string, and stringify when creating the error?  Or should I carry the `Value`-typed limit all the way through, and pass that to the error constructor?  ("Should" in the sense of, what best matches the approach taken elsewhere in this codebase)

djmitche:
Well, I [gave that a try](https://github.com/Stranger6667/jsonschema-rs/compare/master...djmitche:issue215?expand=1) and it didn't work out too well:

```
thread 'error::tests::limit_error_display::_number_from_f64_3_0_unwrap_3_0_' panicked at 'assertion failed: `(left == right)`
  left: `"42 is greater than or equal to the maximum of 3"`,
 right: `"42 is greater than or equal to the maximum of 3.0"`', src/error.rs:1023:9
```

it turns out `Number` (which is what I meant when I wrote `Value` above) is "smart' about formatting whole numbers.  Digging in, it looks like `Display::fmt(&f64, ..)` is smart about this.  It looks like `impl Display for Value` handles this by re-serializing the value, and gets the `.0`.  So perhaps I just need to use `{:?}` to print the limit, or perhaps I need to pass the full `Value` after all..

djmitche:
OK, [this](https://github.com/Stranger6667/jsonschema-rs/compare/master...djmitche:issue215?expand=1) passes the updated tests; it keeps a `limit_val` as a means to reconstitute the limit representation.  If this is OK with you, I'll do the same for the other limits.

[...]

PR #275, opened 2021-08-14
fix: Show limit errors in their original format

djmitche:
This means that a limit such as `{minimum: 3.0}` will display as `3.0`
in the error message, and not `3`.


djmitche:
Fixes #215.

PR #280, opened 2021-10-02
Replace `ValidationError::schema` with custom errors

### LINKED ISSUE #235: Replace `ValidationError::schema` calls with proper types
Once #226 is merged

zhiburt:
Hi there, bump into here via hacktoberfest mark.

Honestly most likely it doesn't address `JSONPointer` correctly in errors.
And I was not sure which type of error better use for things like as64 error or regex syntax error but I did used `ValidationError::format`.

Note: there might be an issue with cargo fmt.

closes #235

Thank you for the library.

### DIFF
--- a/jsonschema/src/keywords/additional_items.rs
+++ b/jsonschema/src/keywords/additional_items.rs
@@ -1,11 +1,4 @@
-use crate::{
-    compilation::{compile_validators, context::CompilationContext, JSONSchema},
-    error::{error, no_error, ErrorIterator, ValidationError},
-    keywords::{boolean::FalseValidator, CompilationResult},
-    paths::{InstancePath, JSONPointer},
-    schema_node::SchemaNode,
-    validator::{format_validators, Validate},
-};
+use crate::{compilation::{compile_validators, context::CompilationContext, JSONSchema}, error::{error, no_error, ErrorIterator, ValidationError}, keywords::{boolean::FalseValidator, CompilationResult}, paths::{InstancePath, JSONPointer}, primitive_type::{PrimitiveType, PrimitiveTypesBitMap}, schema_node::SchemaNode, validator::{format_validators, Validate}};
 use serde_json::{Map, Value};
 
 pub(crate) struct AdditionalItemsObjectValidator {
@@ -153,7 +146,14 @@ pub(crate) fn compile<'a>(
                     Some(FalseValidator::compile(schema_path))
                 }
             }
-            _ => Some(Err(ValidationError::schema(schema))),
+            _ => Some(Err(ValidationError::multiple_type_error(
+                JSONPointer::default(),
+                context.clone().into_pointer(),
+                schema,
+                PrimitiveTypesBitMap::new()
+                    .add_type(PrimitiveType::Boolean)
+                    .add_type(PrimitiveType::Object),
+            ))),
         }
     } else {
         None
--- a/jsonschema/src/keywords/additional_items.rs
+++ b/jsonschema/src/keywords/additional_items.rs
@@ -1,4 +1,12 @@
-use crate::{compilation::{compile_validators, context::CompilationContext, JSONSchema}, error::{error, no_error, ErrorIterator, ValidationError}, keywords::{boolean::FalseValidator, CompilationResult}, paths::{InstancePath, JSONPointer}, primitive_type::{PrimitiveType, PrimitiveTypesBitMap}, schema_node::SchemaNode, validator::{format_validators, Validate}};
+use crate::{
+    compilation::{compile_validators, context::CompilationContext, JSONSchema},
+    error::{error, no_error, ErrorIterator, ValidationError},
+    keywords::{boolean::FalseValidator, CompilationResult},
+    paths::{InstancePath, JSONPointer},
+    primitive_type::{PrimitiveType, PrimitiveTypesBitMap},
+    schema_node::SchemaNode,
+    validator::{format_validators, Validate},
+};
 use serde_json::{Map, Value};
 
 pub(crate) struct AdditionalItemsObjectValidator {
--- a/jsonschema/src/keywords/additional_items.rs
+++ b/jsonschema/src/keywords/additional_items.rs
@@ -159,8 +159,9 @@ pub(crate) fn compile<'a>(
                 context.clone().into_pointer(),
                 schema,
                 PrimitiveTypesBitMap::new()
-                    .add_type(PrimitiveType::Boolean)
-                    .add_type(PrimitiveType::Object),
+                    .add_type(PrimitiveType::Object)
+                    .add_type(PrimitiveType::Array)
+                    .add_type(PrimitiveType::Boolean),
             ))),
         }
     } else {
--- a/jsonschema/src/keywords/additional_properties.rs
+++ b/jsonschema/src/keywords/additional_properties.rs
@@ -1364,7 +1364,7 @@ fn compile_patterns<'a>(
                 JSONPointer::default(),
                 keyword_context.clone().into_pointer(),
                 subschema,
-                "patternProperties",
+                "regex",
             ));
         }
     }
--- a/jsonschema/src/keywords/additional_properties.rs
+++ b/jsonschema/src/keywords/additional_properties.rs
@@ -1361,7 +1361,7 @@ fn compile_patterns<'a>(
             compiled_patterns.push((compiled_pattern, node));
         } else {
             return Err(ValidationError::format(
-                context.clone().into_pointer(),
+                JSONPointer::default(),
                 keyword_context.clone().into_pointer(),
                 subschema,
                 "patternProperties",
--- a/jsonschema/src/keywords/additional_properties.rs
+++ b/jsonschema/src/keywords/additional_properties.rs
@@ -1357,13 +1357,15 @@ fn compile_patterns<'a>(
     for (pattern, subschema) in obj {
         let pattern_context = keyword_context.with_path(pattern.to_string());
         if let Ok(compiled_pattern) = Regex::new(pattern) {
-            if let Ok(node) = compile_validators(subschema, &pattern_context) {
-                compiled_patterns.push((compiled_pattern, node));
-            } else {
-                return Err(ValidationError::schema(subschema));
-            }
+            let node = compile_validators(subschema, &pattern_context)?;
+            compiled_patterns.push((compiled_pattern, node));
         } else {
-            return Err(ValidationError::schema(subschema));
+            return Err(ValidationError::format(
+                context.clone().into_pointer(),
+                keyword_context.clone().into_pointer(),
+                subschema,
+                "patternProperties",
+            ));
         }
     }
     Ok(compiled_patterns)
--- a/jsonschema/src/keywords/contains.rs
+++ b/jsonschema/src/keywords/contains.rs
@@ -433,19 +433,34 @@ pub(crate) fn compile<'a>(
                                 max_contains,
                             ))
                         } else {
-                            Some(Err(ValidationError::schema(schema)))
+                            Some(Err(ValidationError::format(
+                                context.clone().into_pointer(),
+                                context.clone().into_pointer(),
+                                schema,
+                                "maxContains",
+                            )))
                         }
                     } else {
                         Some(MinContainsValidator::compile(schema, context, min_contains))
                     }
                 } else {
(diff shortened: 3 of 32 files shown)


Stranger6667:
Thank you @zhiburt !

I am happy to see your contribution :) I'll try to review this in the next few days :) Could you, please:

- Run `cargo fmt`
- Remove the unused function
- Add a changelog entry

Stranger6667:
> Honestly most likely it doesn't address JSONPointer correctly in errors.

I think it is alright, in any case, it is an improvement over the status quo

> And I was not sure which type of error better use for things like as64 error or regex syntax error but I did used

That's alright :)

jsonschema/src/keywords/additional_properties.rs
```diff
@@ -1357,13 +1357,15 @@ fn compile_patterns<'a>(
...
-                compiled_patterns.push((compiled_pattern, node));
-            } else {
-                return Err(ValidationError::schema(subschema));
-            }
+            let node = compile_validators(subschema, &pattern_context)?;
+            compiled_patterns.push((compiled_pattern, node));
         } else {
-            return Err(ValidationError::schema(subschema));
+            return Err(ValidationError::format(
+                context.clone().into_pointer(),
```
Stranger6667, line 1364: Generally, we can't point at what part of the JSON meta-schema this error is happening. The context only contains the "instance_path" part for this case.

Stranger6667 (changes requested):
During the compilation phase, `schema_path` means a path within the JSON Schema [meta schemas](https://github.com/Stranger6667/jsonschema-rs/tree/master/jsonschema/meta_schemas), and `instance_path` is a path within the input schema. Given that, the `schema_path` argument to `ValidationError` constructors can't be easily derived, so we can leave it empty - this is a corner case anyway (something that metaschema validation doesn't catch).

So, we need to put pointers constructed from `context` to `instance_path` argument and `JSONPointer::default()` to the `schema_path` one. As far as I see, it is the other way around now.

zhiburt:
> During the compilation phase, schema_path means a path within the JSON Schema meta schemas

So its essentially pointers to this map?

https://github.com/Stranger6667/jsonschema-rs/blob/aaadd99b2cc376d9018759ecedbbe68c0700d2ab/jsonschema/src/compilation/options.rs#L28-L74 

I am curious if we could reuse `self.schema_path` in structures like this as `schema_path` arg in error.

https://github.com/Stranger6667/jsonschema-rs/blob/fba763fd4fed01ae8073c6a0c9204cb8c7080933/jsonschema/src/keywords/pattern.rs#L19-L23

Stranger6667:
> So its essentially pointers to this map?

Yep, I'd say that they point to some locations within the values of this map. 

> I am curious if we could reuse self.schema_path in structures like this as schema_path arg in error.

In the majority of the cases, these `schema_path` values [are already used](https://github.com/Stranger6667/jsonschema-rs/blob/master/jsonschema/src/compilation/options.rs#L197) by the meta-schema validators. If the meta-schema is not capable to catch the bug, then these manual error-handling scenarios should catch the problem. 

Of course, we may try to keep the meta-schema context manually and trace all the paths during the compilation. However, it will be hard to do as we'll need to do it differently for different JSON Schema drafts, so I assume it will be easier to tweak the meta-schema instead - it [happened](https://github.com/Stranger6667/jsonschema-rs/commit/7d8364ec9c65816182c1e2ea3f807a15e9a0716a) before and was caused by an actual bug in meta schemas.

Hope I helps :)

zhiburt:
> I am curious if we could reuse self.schema_path in structures like this as schema_path arg in error.
> > In the majority of the cases, these schema_path values are already used by the meta-schema validators. If the meta-schema is not capable to catch the bug, then these manual error-handling scenarios should catch the problem.

I guess I did make a typo.
I meant to say maybe we could use this `self.schema_path` in errors.
Like this

```rust
ValidationError::format(self.schema_path, context.clone().into_pointer(), ...)
```

jsonschema/src/keywords/contains.rs
```diff
@@ -433,19 +433,34 @@ pub(crate) fn compile<'a>(
                                 max_contains,
                             ))
                         } else {
-                            Some(Err(ValidationError::schema(schema)))
+                            Some(Err(ValidationError::format(
```
Stranger6667, line 436: Here we can check whether the value is `i64` - if so, then it is a negative value (otherwise it should be castable to `u64` in the code above). In this case, `ValidationError::minimum`  should cover our need, otherwise - `ValidationError::single_type_error`. 

This comment should be applicable to other cases in this file & similar cases for `maxItems`, `maxLength`, `maxProperties`, `minItems`, `minLength`, and `minProperties`

jsonschema/src/keywords/max_items.rs
```diff
@@ -18,7 +18,12 @@ impl MaxItemsValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MaxItemsValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

jsonschema/src/keywords/max_length.rs
```diff
@@ -18,7 +18,12 @@ impl MaxLengthValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MaxLengthValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

jsonschema/src/keywords/max_properties.rs
```diff
@@ -18,7 +18,12 @@ impl MaxPropertiesValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MaxPropertiesValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

jsonschema/src/keywords/min_items.rs
```diff
@@ -18,7 +18,12 @@ impl MinItemsValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MinItemsValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

jsonschema/src/keywords/min_length.rs
```diff
@@ -18,7 +18,12 @@ impl MinLengthValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MinLengthValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

jsonschema/src/keywords/min_properties.rs
```diff
@@ -18,7 +18,12 @@ impl MinPropertiesValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MinPropertiesValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                schema_path,
```
Stranger6667, line 22: `schema_path` and `instance_path` should be switched here

Stranger6667 (changes requested):
Added a few comments - most of them are quite similar to each other and mostly are about choosing the proper `ValidationError` variant

jsonschema/src/keywords/required.rs
```diff
@@ -141,16 +149,27 @@ pub(crate) fn compile_with_path(
     match schema {
         Value::Array(items) => {
             if items.len() == 1 {
-                if let Some(Value::String(item)) = items.iter().next() {
+                let item = items.iter().next().unwrap();
```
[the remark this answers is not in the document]
zhiburt: Note: I explicitly used unwrap because it must never happen, because we check len() before.

jsonschema/src/keywords/additional_items.rs
```diff
@@ -153,7 +154,14 @@ pub(crate) fn compile<'a>(
...
                 }
             }
-            _ => Some(Err(ValidationError::schema(schema))),
+            _ => Some(Err(ValidationError::multiple_type_error(
+                JSONPointer::default(),
+                context.clone().into_pointer(),
+                schema,
+                PrimitiveTypesBitMap::new()
+                    .add_type(PrimitiveType::Boolean)
+                    .add_type(PrimitiveType::Object),
```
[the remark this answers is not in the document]
zhiburt: Good catch

jsonschema/src/keywords/contains.rs
zhiburt: > Here we can check whether the value is i64 - if so, then it is a negative value

Its not necessarily true, e.g `1` is a valid `i64` but its not negative. So it would require `< 0` check.

jsonschema/src/keywords/required.rs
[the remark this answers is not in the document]
Stranger6667: Yep, my main motivation is that if in the future the check above will change, and a failure occur, then it will be easier to understand what is going on with `expect`

jsonschema/src/keywords/contains.rs
Stranger6667: If it is `1`, then the execution will go in the branch above, but if it is in this branch, then we exclude all values that are `u64` AND `i64`, so only negative values are possible

zhiburt: Check how I did implement it.

https://github.com/Stranger6667/jsonschema-rs/pull/280/files#diff-96d9e0c98ab5e4b39ff4c741e1ab1958b48e5aa4688f4e0d42e420a2cc2dedabR449-R474

jsonschema/src/keywords/required.rs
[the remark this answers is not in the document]
zhiburt: I actually followed other approach just used index. See the changes.

zhiburt:
You've spend quite a bit of time to review it.

jsonschema/src/keywords/min_properties.rs
```diff
@@ -18,7 +18,12 @@ impl MinPropertiesValidator {
         if let Some(limit) = schema.as_u64() {
             Ok(Box::new(MinPropertiesValidator { limit, schema_path }))
         } else {
-            Err(ValidationError::schema(schema))
+            Err(ValidationError::format(
+                JSONPointer::default(),
+                JSONPointer::default(),
```
Stranger6667, line 23: It should be `schema_path`

jsonschema/src/keywords/contains.rs
```diff
@@ -422,38 +423,55 @@ pub(crate) fn compile<'a>(
...
 
+#[cfg(feature = "draft201909")]
+#[inline]
+pub(crate) fn map_get_u64<'a>(
+    m: &'a Map<String, Value>,
+    context: &CompilationContext,
+    type_name: &str,
+) -> Option<Result<u64, ValidationError<'a>>> {
+    let value = m.get(type_name)?;
+    if matches!(value.as_i64(), Some(n) if n < 0) {
```
Stranger6667, line 455: I really like that you factored out this logic into a separate function :) Maybe it could also be moved to the `helpers.rs` file to reuse it in other keywords.

Though, there are two calls to `as_<type>` functions unconditionally, the previous approach would have only one in the valid input scenario.

It would be nice to move this block to the `None` branch of the `match` statement below to keep the happy path simpler. Then, the `n < 0` check will not be needed as well - it could be as simple as `value.is_i64()` (as all non-negative values will be covered by `as_u64()`)

Stranger6667 (changes requested):
Thanks! I think we're getting closer to merging :) A couple of minor things

jsonschema/src/keywords/contains.rs
zhiburt: Actually you're right about 2 calls. :100:

Stranger6667:
Awesome! Thanks :)

PR #281, opened 2021-10-02
Disable format validation

### LINKED ISSUE #261: Option to turn off processing of the `format` keyword
From the Draft 2019-09 [release notes](https://json-schema.org/draft/2019-09/release-notes.html):

> The key difference here is that format validation is now predictably off by default, but can be configured to be turned on. In draft-07, it was on (but possibly unimplemented) by default and could be configured to be turned off.

The release notes page contains a table with possible validation outcomes, depending on multiple factors - `support level`, `configuration`, and `vocabulary`. In our case, we don't support vocabularies just yet and I don't see much reason to distinguish "full syntax" vs "best effort" support level, so I think we can reduce that table to this one:

| supported        | configuration           | outcome  |
| -------------: | -------------:| -----:|
| no      | - | not validated |
| yes      | default (off) | not validated |
| yes      | on | validated |

When vocabularies will be implemented, we can revisit this behavior and add some `VocabularyError` to the mix.

On the implementation side, it can be controlled with a new configuration value in `compilation::options::CompilationOptions` that will turn format validation on and off (with `false` as the default value in `Default` impl). Let's say it is `validate_formats`, then its value is `false` on the new draft (which is behind a feature flag at the moment) and `true` on older drafts to maintain backward-compatible behavior.

This change will require a setter on `compilation::options::CompilationOptions` similarly to other setters there.
Then, depending on this value we can just return `None` from `keywords::format::compile` if it is `false`, which will turn format validation off.

jacobmischka:
Prior to draft versions 2019-09, format validation is enabled by default (though should be able to be disabled), and vice versa afterword. This patch adds `should_validate_formats` to CompilationOptions to force the option one way or another.

If not specified, it will fallback to a default based on the draft version (enabled by default before draft 2019-09).

I also rolled a small typo fix into this PR in a separate commit, let me know if you'd like me to squash the commits, or if you'd like that as a separate PR, or something else.

Closes #261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - `uuid` format validator. [#266](https://github.com/Stranger6667/jsonschema-rs/issues/266)
 - `duration` format validator. [#265](https://github.com/Stranger6667/jsonschema-rs/issues/265)
 - Collect annotations whilst evaulating schemas.[#262](https://github.com/Stranger6667/jsonschema-rs/issues/262)
+- Option to turn off processing of the `format` keyword. [#261](https://github.com/Stranger6667/jsonschema-rs/issues/261)
 - `basic` & `flag` output formatting styles. [#100](https://github.com/Stranger6667/jsonschema-rs/issues/100)
 
 ### Changed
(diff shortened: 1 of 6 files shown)


Stranger6667:
Awesome!

Thank you @jacobmischka for your contribution! 

Please:
- Add a changelog entry
- Fix clippy warnings
- Add a small test that formats are ignored when `false` is passed

Besides these minor things, the PR looks great to me!

> I also rolled a small typo fix into this PR in a separate commit, let me know if you'd like me to squash the commits, or if you'd like that as a separate PR, or something else.

Thanks :) Let's leave it as it is :)

Stranger6667 (approved):

Stranger6667:
Thanks!

Issue #285, opened 2021-10-03
Reduce the size of `JSONPointer`


Stranger6667:
Similar to #284, the underlying `Vec<PathChunk>` could be replaced with `Box<[PathChunk]>` - it will require some changes in the way the pointer is extended during the compilation phase. However, the underlying vector is still cloned in any case.

It should help create `ValidationError` instances, as most of its constructors receive two `JSONPointer`, and it will save 16 bytes when each error is created, which should be visible. Also, `ValidationError` will be smaller by 16 bytes;

Stranger6667:
Indeed, I think it won’t give significant performance improvement

PR #293, opened 2021-10-09
perf: Reduce the size of JSONPointer

pdogr:
Fixes: #285 
Changes
- Replace ```Vec<PathChunk>``` with ```Box<[PathChunk]>```.
- ```JSONPointer::clone_with``` and ```JSONPointer::extend_with```  convert the underlying slice to vector before extending. Not an issue as the vector was being cloned already. 
- ```fn into_iter(self) -> Self::IntoIter``` consumes the underlying slice via ```into_vec```.

### DIFF
--- a/jsonschema/src/paths.rs
+++ b/jsonschema/src/paths.rs
@@ -3,13 +3,14 @@ use std::{fmt, fmt::Write, slice::Iter, str::FromStr};
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 /// JSON Pointer as a wrapper around individual path components.
-pub struct JSONPointer(Vec<PathChunk>);
+pub struct JSONPointer(Box<[PathChunk]>);
 
 impl JSONPointer {
     #[must_use]
     /// JSON pointer as a vector of strings. Each component is casted to `String`. Consumes `JSONPointer`.
     pub fn into_vec(self) -> Vec<String> {
         self.0
+            .into_vec()
             .into_iter()
             .map(|item| match item {
                 PathChunk::Property(value) => value.into_string(),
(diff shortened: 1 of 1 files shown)


Stranger6667:
Thanks! Answering your [comment](https://github.com/Stranger6667/jsonschema-rs/issues/285#issuecomment-939262100):

> Hi, I tried to fix this in #293, but I think that this might lead to a decrease in performance as currently I am using into_vec to consume the Box<[PathChunk]> in the iterator. Please do let me know what you think.
Thanks!

I think that there is a trade-off - the `IntoIter` implementation is not used in the lib itself, so it is for public use only and it will slightly decrease its performance. On the other hand, `ValidationError` instances are created during the validation process and users use them later. 

My assumption is that saving 16 bytes on each `ValidationError` (as it requires two instances of `JSONPointer`) might be more important than a slowdown on `IntoIter`. But, it would be better to measure it - `cargo bench --bench jsonschema "invalid"` will include cases when `ValidationError` instances are created. I am surely can be wrong about my assumptions :)

The benchmark suite might not have the best naming of the cases, so it might be hard to select the right ones. But the representative ones ends with `jsonschema_rs/validate` and have `invalid` in its names (e.g. `additional_items_boolean invalid 1234/jsonschema_rs/validate`). It also might be easier to change the order in benchmark names so it will be easier to select the right ones

Stranger6667:
As an extra thing - boxed slices might slow down the validators compilation phase as well, but the runtime performance is more important

Stranger6667:
I'll unify benchmarks name, so running `cargo bench --bench jsonschema "jsonschema_rs/validate/invalid"` will be enough

PR #296, opened 2021-10-19
feat(python): support iterating all errors with `iter_errors`

### LINKED ISSUE #236: Python: Add an iterator over validation errors
Now only the first error is taken. It could be ‘iter_errors’ function / method

blacha:
After fighting with the `'static` lifetime for more time that I'd like to admit, I thought that it would be easy to just have the errors returned as `PyList`

This PR adds a python function `list_errors()` and `compiled.list_errors()` that returns a `PyList<PyErr>` of all the validation errors or a empty list if there are no validation errors.

I am a rust newbie, so any feedback is appreciated. 

Fixes #236


Stranger6667:
> I feel like the 'static might not be a good idea anyway, seems like a recipe for leaking memory..
> I have switched this across to the vec implementation you suggested! thanks for that idea.

You're very welcome! 

The only lifetime inside `jsonschema::ErrorIterator` goes to `Cow<'a, Value>`, which in the case of `'static` means that the `Cow::Owned` variant is used and it should be alright. Nevertheless, the more I think about it, the more I realize that error types without lifetimes are the most flexible ones :(

blacha:
> I have a few minor comments :) Otherwise, looks good to me

I believe I have resolved all of your comments.

PR #312, opened 2021-11-01
fix: convert python tuples into lists

blacha:
We use tuples for geographic coordinates and bounding boxes which doesn't seem to have a direct mapping in JSON Schema.

This pull requests converts the tuple into a list to fix `ValueError: Unsupported type: 'tuple'` when validating a object that contains a tuple.

```
{ "bbox": (1,2,3,4) }
```

### DIFF
--- a/bindings/python/src/ser.rs
+++ b/bindings/python/src/ser.rs
@@ -2,7 +2,7 @@ use pyo3::{
     exceptions,
     ffi::{
         PyDictObject, PyFloat_AS_DOUBLE, PyList_GET_ITEM, PyList_GET_SIZE, PyLong_AsLongLong,
-        Py_TYPE, PyTuple_GET_ITEM, PyTuple_GET_SIZE
+        PyTuple_GET_ITEM, PyTuple_GET_SIZE, Py_TYPE,
     },
     prelude::*,
     types::PyAny,
@@ -149,11 +149,16 @@ impl Serialize for SerializePyObject {
                     map.end()
                 }
             }
-            ObjectType::List => {
+            ObjectType::Tuple | ObjectType::List => {
                 if self.recursion_depth == RECURSION_LIMIT {
                     return Err(ser::Error::custom("Recursion limit reached"));
                 }
-                let length = unsafe { PyList_GET_SIZE(self.object) } as usize;
+
+                let length = match self.object_type {
+                    ObjectType::Tuple => unsafe { PyTuple_GET_SIZE(self.object) as usize },
+                    ObjectType::List => unsafe { PyList_GET_SIZE(self.object) as usize },
+                    _ => return Err(ser::Error::custom("Object is not a list or tuple")),
+                };
                 if length == 0 {
                     serializer.serialize_seq(Some(0))?.end()
                 } else {
@@ -161,35 +166,13 @@ impl Serialize for SerializePyObject {
                     let mut ob_type = ObjectType::Str;
                     let mut sequence = serializer.serialize_seq(Some(length))?;
                     for i in 0..length {
-                        let elem = unsafe { PyList_GET_ITEM(self.object, i as isize) };
-                        let current_ob_type = unsafe { Py_TYPE(elem) };
-                        if current_ob_type != type_ptr {
-                            type_ptr = current_ob_type;
-                            ob_type = get_object_type(current_ob_type);
-                        }
-                        #[allow(clippy::integer_arithmetic)]
-                        sequence.serialize_element(&SerializePyObject::with_obtype(
-                            elem,
-                            ob_type.clone(),
-                            self.recursion_depth + 1,
-                        ))?;
-                    }
-                    sequence.end()
-                }
-            }
-            ObjectType::Tuple => {
-                if self.recursion_depth == RECURSION_LIMIT {
-                    return Err(ser::Error::custom("Recursion limit reached"));
-                }
-                let length = unsafe { PyTuple_GET_SIZE(self.object) } as usize;
-                if length == 0 {
-                    serializer.serialize_seq(Some(0))?.end()
-                } else {
-                    let mut type_ptr = std::ptr::null_mut();
-                    let mut ob_type = ObjectType::Str;
-                    let mut sequence = serializer.serialize_seq(Some(length))?;
-                    for i in 0..length {
-                        let elem = unsafe { PyTuple_GET_ITEM(self.object, i as isize) };
+                        let elem = match self.object_type {
+                            ObjectType::Tuple => unsafe {
+                                PyTuple_GET_ITEM(self.object, i as isize)
+                            },
+                            ObjectType::List => unsafe { PyList_GET_ITEM(self.object, i as isize) },
+                            _ => return Err(ser::Error::custom("Object is not a list or tuple")),
+                        };
                         let current_ob_type = unsafe { Py_TYPE(elem) };
                         if current_ob_type != type_ptr {
                             type_ptr = current_ob_type;
--- a/bindings/python/src/ser.rs
+++ b/bindings/python/src/ser.rs
@@ -2,7 +2,7 @@ use pyo3::{
     exceptions,
     ffi::{
         PyDictObject, PyFloat_AS_DOUBLE, PyList_GET_ITEM, PyList_GET_SIZE, PyLong_AsLongLong,
-        Py_TYPE,
+        Py_TYPE, PyTuple_GET_ITEM, PyTuple_GET_SIZE
     },
     prelude::*,
     types::PyAny,
@@ -27,6 +27,7 @@ pub enum ObjectType {
     Float,
     List,
     Dict,
+    Tuple,
     Unknown(String),
 }
 
@@ -81,6 +82,8 @@ pub fn get_object_type(object_type: *mut pyo3::ffi::PyTypeObject) -> ObjectType 
         ObjectType::None
     } else if object_type == unsafe { types::LIST_TYPE } {
         ObjectType::List
+    } else if object_type == unsafe { types::TUPLE_TYPE } {
+        ObjectType::Tuple
     } else if object_type == unsafe { types::DICT_TYPE } {
         ObjectType::Dict
     } else {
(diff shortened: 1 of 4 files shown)


bindings/python/src/ser.rs
```diff
@@ -174,6 +177,34 @@ impl Serialize for SerializePyObject {
                     sequence.end()
                 }
             }
+            ObjectType::Tuple => {
```
blacha, line 180: This is copy pasted from the `ObjectType::List` option the only change is `PyTuple_GET_SIZE` vs `PyList_GET_SIZE` and `GET_ITEM`

Is there a better way to merge these into one ?

bindings/python/src/types.rs
```diff
@@ -24,6 +25,7 @@ pub fn init() {
         TRUE = Py_True();
         STR_TYPE = Py_TYPE(PyUnicode_New(0, 255));
         DICT_TYPE = Py_TYPE(PyDict_New());
+        TUPLE_TYPE = Py_TYPE(PyTuple_New(1));
```
blacha, line 28: Should this be `0_isize` too ?

Stranger6667:
Hi @blacha 

Thanks for opening a PR! Though I clearly see the value of this change for your use case, I think it might be a problematic change for the library as a whole. Unfortunately, I didn't share the main design points anywhere, so I'll try to put it here (and will later put it in the docs).

1. As you pointed out, `tuple` doesn't have a direct equivalent in JSON, though it feels like a `list` in a general sense. The core thing here is to avoid implicit type conversions, so it is more conservative by default to avoid surprising behavior.
2. Compatibility with `jsonschema` in its Python-specific idiosyncrasies. `jsonschema` ignores tuples - if somebody will switch from `jsonschema` to `jsonschema_rs` it might be confusing as tuples will be converted to lists & validation will be performed. It is not necessarily a bad thing, as it will uncover that non-JSONSchema types were used.

So, my main intention is to keep the library conservative to its inputs by default so it is less surprising and more explicit about deviations from the JSON Schema spec.

However, I think that we still can add this change to the library but put it behind a feature flag, similarly to what `orjson` is [doing](https://github.com/ijl/orjson/blob/master/src/opt.rs) (feature flags for supporting subclasses, out-of-range int values, etc).

So, the end-user API might look like this:

```python
jsonschema_rs.validate({"items": {"type": "integer"}}, [1, "2"], option=jsonschema_rs.OPT_TUPLE_AS_LIST)

compiled = jsonschema_rs.JSONSchema({"items": {"type": "integer"}}, option=jsonschema_rs.OPT_TUPLE_AS_LIST)
```

What do you think?

blacha:
> Hi @blacha
> 
> Thanks for opening a PR! Though I clearly see the value of this change for your use case, I think it might be a problematic change for the library as a whole. Unfortunately, I didn't share the main design points anywhere, so I'll try to put it here (and will later put it in the docs).
> 
> 1. As you pointed out, `tuple` doesn't have a direct equivalent in JSON, though it feels like a `list` in a general sense. The core thing here is to avoid implicit type conversions, so it is more conservative by default to avoid surprising behavior.
> 2. Compatibility with `jsonschema` in its Python-specific idiosyncrasies. `jsonschema` ignores tuples - if somebody will switch from `jsonschema` to `jsonschema_rs` it might be confusing as tuples will be converted to lists & validation will be performed. It is not necessarily a bad thing, as it will uncover that non-JSONSchema types were used.
> 
> So, my main intention is to keep the library conservative to its inputs by default so it is less surprising and more explicit about deviations from the JSON Schema spec.
> 
> However, I think that we still can add this change to the library but put it behind a feature flag, similarly to what `orjson` is [doing](https://github.com/ijl/orjson/blob/master/src/opt.rs) (feature flags for supporting subclasses, out-of-range int values, etc).
> 
> So, the end-user API might look like this:
> 
> ```python
> jsonschema_rs.validate({"items": {"type": "integer"}}, [1, "2"], option=jsonschema_rs.OPT_TUPLE_AS_LIST)
> 
> compiled = jsonschema_rs.JSONSchema({"items": {"type": "integer"}}, option=jsonschema_rs.OPT_TUPLE_AS_LIST)
> ```
> 
> What do you think?

We are coming from `fastjsonschema` which does this conversion automatically https://github.com/horejsek/python-fastjsonschema/issues/87

I do see that `jsonschema` does follow a similar pattern to the configuration options you are talking about https://github.com/Julian/jsonschema/issues/148 it does have a much more complicated type setting option

I am for the automatic conversion of tuples to arrays, as there is no concept of tuples in JSON. If I was to take a tuple dict and seralize it to JSON using `json.dumps` then parse it as JSON `json.loads` then validate it would work, as the serialisation process automatically converts tuples into arrays

```python
import json
json.dumps({"foo": (1,2,3)})
# '{"foo": [1, 2, 3]}'

validate(json.loads(json.dumps({"foo": (1,2,3)})), schema) # Works
```
It looks somewhat easy to plumb the configuration option into the serialiser via the `ser::to_value` function. 

Interested in your thoughts if its worth the effort to create a config for this

Stranger6667:
Generally, I am somewhat hesitant to follow Python's `json` behavior as it serializes `Inf` / `-Inf` / `nan` which do not exist in JSON (which leads to data that can't be parsed by other languages). But specifically, in this case, I do agree that it would be convenient to have this kind of automatic conversion, especially as the scope is limited to Python bindings.

> It looks somewhat easy to plumb the configuration option into the serialiser via the ser::to_value function.
> Interested in your thoughts if its worth the effort to create a config for this

I think so, yes. The `to_value` function might accept `u8` (a bitmask for enabled features) and then decide during the serialization process whether to convert tuples or not. As it is a private implementation detail, we can use `u8` for now as we don't have many config options and we'll always can extend it to `u16` in the future if needed.

On the public API side, I think it could be the `orjson` [approach](https://github.com/ijl/orjson/blob/master/src/lib.rs#L293) with a single `options` argument (which is also similar to Python's `re` config options). 

Indeed, it should not be a big deal to extend the current public API to accept a new argument, process it, and forward the value to `to_value`.

Let me know if I can help with the implementation!

bindings/python/src/ser.rs
```diff
@@ -146,19 +149,30 @@ impl Serialize for SerializePyObject {
                     map.end()
                 }
             }
-            ObjectType::List => {
+            ObjectType::Tuple | ObjectType::List => {
```
Stranger6667, line 152: We can avoid the `match` statements below (with their unreachable `_` arms) by separating these two arms - the code is somewhat similar, but it could be solved by a macro. This way we'll avoid the performance penalty of checking the type of each list / tuple multiple times.

bindings/python/tests-py/test_jsonschema.py
```diff
@@ -65,6 +65,12 @@ def test_from_str_error():
         JSONSchema.from_str(42)
 
 
+def test_tuple():
+    schema = {"properties": {"foo": {"type": "array"}}}
+    instance = {"foo": (1, 2, 3)}
+    assert is_valid(instance, schema) == True
```
Stranger6667, line 71: As the default behavior for type mismatch is to consider the input valid, it would be nice to have a test case that will examine the situation when the input is not valid.

blacha: with out this change, this test throws:

```
ValueError: Unsupported type: 'tuple'
```

bindings/python/src/ser.rs
blacha: I had this originally as two separate Matches, but wanted to reduce the code duplication, I can't imagine the match function is super slow in here, I'm assuming the python calls will be the slowest part here by a large margin

I've never written a rust macro, so will do some reading on how they work.

bindings/python/tests-py/test_jsonschema.py
blacha: Ahh I now see what you mean, `is_valid({}, schema) is True` weird, Ill fix this.

blacha:
@Stranger6667 Playing with `orjson` it automaticallys convert tuples by default

```python
import orjson

print(orjson.dumps({"foo": (1,2,3)}))
print(orjson.dumps({"foo": [1,2,3]}))
```

It only avoids the named tuple type

```python
from collections import namedtuple
Person = namedtuple('Person', 'first_name last_name')
person_a = Person('Joe', 'Smith')

print(orjson.dumps(person_a)) # Dies with TypeError: Type is not JSON serializable: Person
```

Which makes me think it should just by default convert, I'll add a test for named tuples.

bindings/python/tests-py/test_jsonschema.py
```diff
@@ -87,6 +88,14 @@ def test_array_tuple_invalid(val):
         validate(schema, val)
 
 
+def test_named_tuple():
+    Person = namedtuple("Person", "first_name last_name")
+    person_a = Person("Joe", "Smith")
+    schema = {"type": "array", "items": {"type": "string"}}
+    with pytest.raises(ValueError):
```
blacha, line 95: `namedtuple`'s error with `ValueError: Unsupported type: 'Person'`

Stranger6667: Great!

bindings/python/src/ser.rs
Stranger6667: Agree, the overhead should be relatively small here, but:

- It scales together with the number of lists & number of elements in each list, so it would be more visible with larger list/tuple heavy inputs;
- I assume that it also affects cache locality a bit as we always need to keep one more value around (`self.object_type`)
- It forces us to write code for unreachable cases;

The performance points need some profiling of course but from the top level, I prefer to avoid performance costs that have a relatively straightforward solution.

I am totally ok to have a bit of code duplication here as the blocks are relatively small.

blacha: I did have a read through some tutorials on macros, but am a little lost on how to actually implement a macro here.

I could just copy and paste the code into match blocks if that would work for you, or if you have any input on how to make a macro for this, I'd be happy to update the PR.

Stranger6667:
@blacha 

Good point! I didn't know that `orjson` does it by default. Lets make it the default behavior then :)

bindings/python/src/ser.rs
Stranger6667: > I could just copy and paste the code into match blocks if that would work for you, or if you have any input on how to make a macro for this, I'd be happy to update the PR.

Absolutely! We can think about a macro later

Stranger6667:
Also, we might get into the bitmask options later :)

blacha:
> Also, we might get into the bitmask options later :)

haha, I enjoy reverse engineering old games, so I love it when I get a chance to put a bitmask somewhere :) 

however if we just go for the default for tuples as lists I'm not really seeing a place where we need to turn that off, so to keep this PR somewhat small maybe the options logic could be added later?

Stranger6667:
> haha, I enjoy reverse engineering old games, so I love it when I get a chance to put a bitmask somewhere :)
> however if we just go for the default for tuples as lists I'm not really seeing a place where we need to turn that off, so to keep this PR somewhat small maybe the options logic could be added later?

Yep, exactly :) It would be better to implement the tuple conversion for now and then if it will be needed, we can always get back into implementing those bitmask options later :)

bindings/python/src/ser.rs
blacha: Awesome, I have added a commit that does this 

https://github.com/Stranger6667/jsonschema-rs/pull/312/commits/0ada8f37dfb0bd7d7ff9bc11ba390fffa4c1dfe7

Stranger6667 (approved):

Stranger6667:
Thanks! I'll make a new release soon

PR #313, opened 2021-11-03
feat: expose methods to examine OutputUnits

alexjg:
The `Output::basic` method returns a BasicOutput. Prior to this patch the
only way to examine the contents of this data structure was to serialize
it to JSON. This patch adds methods to examine the outputs from Rust, as
well as documentation on how to do so.

Signed-off-by: Alex Good <alex@memoryandthought.me>


jsonschema/src/output.rs
```diff
@@ -169,11 +231,56 @@ impl<T> OutputUnit<T> {
...
         }
     }
+
+    ///  The location in the schema of the keyword
+    pub const fn keyword_location(&self) -> &JSONPointer {
+        &self.keyword_location
+    }
+
+    ///  The absolute location in the schema of the keyword. This will be
+    ///  different to `keyword_location` if the schema is a resovled reference.
```
Stranger6667, line 241:
```suggestion
    ///  different to `keyword_location` if the schema is a resolved reference.
```

jsonschema/src/output.rs
```diff
@@ -169,11 +231,56 @@ impl<T> OutputUnit<T> {
...
 pub struct Annotations<'a>(AnnotationsInner<'a>);
 
+impl<'a> Annotations<'a> {
+    /// The `serde_json::Value` of the annotation
+    pub fn value(&'a self) -> Cow<'a, serde_json::Value> {
+        match &self.0 {
+            AnnotationsInner::Value(v) => Cow::Borrowed(v),
+            AnnotationsInner::ValueRef(v) => Cow::Borrowed(v),
+            AnnotationsInner::UnmatchedKeywords(kvs) => {
+                let value = serde_json::to_value(kvs).unwrap();
```
Stranger6667, line 277: Could we use `expect` here instead? Not sure what would be the situation when panic may happen

Stranger6667 (changes requested):
Thank you, that's awesome! I have two minor comments :)

jsonschema/src/output.rs
alexjg: Good spot, fixed.

jsonschema/src/output.rs
alexjg: I've added an `expect`. The error cases mentioned in `serde_json::to_value` are:

> This conversion can fail if `T`'s implementation of `Serialize` decides to fail, or if `T` contains a map with non-string keys.

But `T` in  our case is `AHashMap<String, serde_json::Value>`, which does not fall into either of these cases.

Stranger6667: Thanks!

Stranger6667 (approved):

Stranger6667:
Released in 0.13.2!

PR #317, opened 2021-11-08
feat: convert python enum into raw values before validating

### LINKED ISSUE #316: Python enum validation
Hi @Stranger6667, Another issue that I'm running into is with Python `enum`'s

Given the following piece of code by default the `Enum` is not converted to a string

```python
from jsonschema_rs import is_valid
import orjson

from enum import Enum

class Foo(Enum):
   bar = "bar"
   foo = "foo"

schema = {"properties": {"foo": {"type": "string"}}}
instance = {"foo": Foo.bar}

converted = orjson.loads(orjson.dumps(instance))
print(is_valid(schema, converted)) # True

print(is_valid(schema, instance)) # ValueError: Unsupported type: 'Foo'
```

I am thinking of two options and wondering which you would prefer

1. Automatically converting Enums to their corresponding type (str, int, etc) like how orjson does it
2. Creating a `string` based interface `validate_str(orjson.dumps(instance))` 

I like the simplicity of 1 for the end user as they don't have to worry about converting their enum's with `Foo.bar.value` before validating.

blacha:
Fixes #316

### DIFF
--- a/bindings/python/src/ffi.rs
+++ b/bindings/python/src/ffi.rs
@@ -2,8 +2,8 @@ use std::os::raw::c_char;
 
 /// ob_type is not exposed by default as is needed to check if something inherits from a enum
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
-pub struct PyTypeObjectSubObType {
+#[derive(Debug)]
+pub struct PyTypeObject {
     pub ob_refcnt: pyo3::ffi::Py_ssize_t,
     pub ob_type: *mut pyo3::ffi::PyTypeObject,
     pub tp_name: *const c_char,
--- a/bindings/python/src/ffi.rs
+++ b/bindings/python/src/ffi.rs
@@ -1,0 +1,10 @@
+use std::os::raw::c_char;
+
+/// ob_type is not exposed by default as is needed to check if something inherits from a enum
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct PyTypeObjectSubObType {
+    pub ob_refcnt: pyo3::ffi::Py_ssize_t,
+    pub ob_type: *mut pyo3::ffi::PyTypeObject,
+    pub tp_name: *const c_char,
+}
--- a/bindings/python/src/ser.rs
+++ b/bindings/python/src/ser.rs
@@ -62,6 +62,13 @@ impl SerializePyObject {
     }
 }
 
+#[inline]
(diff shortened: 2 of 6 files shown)


bindings/python/src/ffi.rs
```diff
@@ -0,0 +1,10 @@
+use std::os::raw::c_char;
+
+/// ob_type is not exposed by default as is needed to check if something inherits from a enum
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct PyTypeObjectSubObType {
```
Stranger6667, line 6: As it corresponds to the `PyTypeObject` struct in CPython, it might be better to keep the original name, so we know that it is our "re-definition" (because the one in PyO3 is opaque and we can't get the `ob_type` field from it)

bindings/python/src/ffi.rs
```diff
@@ -0,0 +1,10 @@
+use std::os::raw::c_char;
+
+/// ob_type is not exposed by default as is needed to check if something inherits from a enum
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
```
Stranger6667, line 5: I assume we don't need `Copy` & `Clone` impls? If not, then I suggest removing them

bindings/python/src/ser.rs
```diff
@@ -86,6 +87,10 @@ pub fn get_object_type(object_type: *mut pyo3::ffi::PyTypeObject) -> ObjectType
         ObjectType::Tuple
     } else if object_type == unsafe { types::DICT_TYPE } {
         ObjectType::Dict
+    } else if unsafe { (*(object_type as *mut ffi::PyTypeObjectSubObType)).ob_type }
```
Stranger6667, line 90: Maybe we can put this check into a separate function (with `#[inline]`)? E.g. `is_enum_subclass`

Stranger6667 (changes requested):
I have a few small readability/naming comments, otherwise, it looks great to me!

Thank you for implementing this :)

bindings/python/src/ffi.rs
blacha: I originally had it named this way but changed the name so I didnt have two `PyTypeObject` (Kinda confusing as to which one you are using) will switch it back

Stranger6667 (approved):
Thanks! Looks good to me! I'll make a new release tomorrow :)

Stranger6667:
Released in [0.13.1](https://pypi.org/project/jsonschema-rs/0.13.1/) :tada:

PR #341, opened 2022-01-28
feat: Custom Resolvers for external references

tamasfe:
A PR for #246. It does not touch on the `async` feature, but makes the library build on `wasm32-unknown-unknown` with `default-features = false` and enables my use-case in [taplo](https://github.com/tamasfe/taplo).

### Added

The `SchemaResolver` trait that is supposed to resolve external schemas by given URLs along with a `DefaultResolver` that does what was already implemented, and additionally also resolves `file` schemes via std.

I decided to just use `Arc<dyn SchemaResolver>` as it is only used by ref validators currently, and adding generics everywhere would've added too much complexity.

A `ValidationError::Resolver` variant was also added accordingly to report if any of the external resolutions failed.

A compiler option to set a custom resolver was also added.

The following features were added to customize the default resolver:

- `resolve-http`: blocking reqwest
- `resolve-file`: resolution via `std::fs`

Both of them are enabled by default so there should be no breaking changes for current users.

### Removed

The reqwest error variant.

All `reqwest-*` features, the users of the library can add `reqwest` as a dependency itself and enable the features they wish, this is a breaking change, although I don't know if anyone used these features.

### Additional Dependencies

- **anyhow**: It's very commonly used, and is somewhat better than bare `Box<dyn std::error::Error>`, although it's not strictly required for this PR.

### TODO

More documentation and tests perhaps?

Also I haven't yet ran or compared the benchmarks, but I don't think that there would be any visible performance hits.

### DIFF
--- a/jsonschema/Cargo.toml
+++ b/jsonschema/Cargo.toml
@@ -17,19 +17,16 @@ name = "jsonschema"
 
 [features]
 cli = ["structopt"]
-default = ["reqwest", "cli"]
+default = ["resolve-http", "resolve-file", "cli"]
 draft201909 = []
 draft202012 = []
-reqwest-native-tls = ["reqwest/native-tls"]
-reqwest-native-tls-alpn = ["reqwest/native-tls-alpn"]
-reqwest-native-tls-vendored = ["reqwest/native-tls-vendored"]
-reqwest-rustls-tls = ["reqwest/rustls-tls"]
-reqwest-rustls-tls-manual-roots = ["reqwest/rustls-tls-manual-roots"]
-reqwest-rustls-tls-native-roots = ["reqwest/rustls-tls-native-roots"]
-reqwest-rustls-tls-webpki-roots = ["reqwest/rustls-tls-webpki-roots"]
+
+resolve-http = ["reqwest"]
+resolve-file = []
 
 [dependencies]
 ahash = { version = "0.7", features = ["serde"] }
+anyhow = "1.0.53"
 base64 = ">= 0.2"
 bytecount = { version = "0.6", features = ["runtime-dispatch-simd"] }
 fancy-regex = "^0.7.1"
@@ -47,7 +44,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 structopt = { version = ">= 0.3", optional = true }
 time = { version = ">= 0.3.3", features = ["parsing", "macros"] }
-url = "2"
+url = "2.2.2"
 uuid = "0.8"
 
 [dev-dependencies]
--- a/jsonschema/Cargo.toml
+++ b/jsonschema/Cargo.toml
@@ -26,7 +26,7 @@ resolve-file = []
 
 [dependencies]
 ahash = { version = "0.7", features = ["serde"] }
-anyhow = "1.0.53"
+anyhow = "1"
 base64 = ">= 0.2"
 bytecount = { version = "0.6", features = ["runtime-dispatch-simd"] }
 fancy-regex = "^0.7.1"
--- a/jsonschema/Cargo.toml
+++ b/jsonschema/Cargo.toml
@@ -44,7 +44,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 structopt = { version = ">= 0.3", optional = true }
 time = { version = ">= 0.3.3", features = ["parsing", "macros"] }
-url = "2.2.2"
+url = "2"
 uuid = "0.8"
 
 [dev-dependencies]
--- a/jsonschema/src/error.rs
+++ b/jsonschema/src/error.rs
@@ -704,14 +704,6 @@ impl<'a> ValidationError<'a> {
             schema_path,
         }
     }
-    pub(crate) fn unknown_reference_scheme(scheme: String) -> ValidationError<'a> {
-        ValidationError {
-            instance_path: JSONPointer::default(),
-            instance: Cow::Owned(Value::Null),
-            kind: ValidationErrorKind::UnknownReferenceScheme { scheme },
-            schema_path: JSONPointer::default(),
-        }
-    }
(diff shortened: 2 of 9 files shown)


jsonschema/Cargo.toml
```diff
@@ -17,19 +17,16 @@ name = "jsonschema"
 
 [features]
 cli = ["structopt"]
-default = ["reqwest", "cli"]
+default = ["resolve-http", "resolve-file", "cli"]
 draft201909 = []
 draft202012 = []
-reqwest-native-tls = ["reqwest/native-tls"]
```
Stranger6667, line 23: As it was asked in #306 I think [this](https://github.com/ahungrynoob/jsonschema/blob/master/Cargo.toml) is the only place where it was relevant. So, with `default-features = false`, these changes should not cause any issues there

jsonschema/src/error.rs
```diff
@@ -143,6 +142,13 @@ pub enum ValidationErrorKind {
     UniqueItems,
     /// Reference contains unknown scheme.
     UnknownReferenceScheme { scheme: String },
+    /// Error during schema ref resolution.
```
Stranger6667, line 145: Should it be behind `cfg` as it was with now removed `Reqwest` variant?

jsonschema/src/resolver.rs
```diff
@@ -11,8 +11,85 @@ use serde_json::Value;
...
+/// A resolver that resolves external schema references.
+/// Internal references such as `#/definitions` and JSON pointers are handled internally.
+///
+/// All operations are blocking and it is not possible to return futures.
+/// As a workaround, errors can be returned that will contain the schema URLs to resolve
+/// and can be resolved outside the validation process if needed.
+///
+/// # Example
+///
+/// ```rust,ignore
```
Stranger6667, line 27: It would be nice if doctests can check whether it compiles or not

jsonschema/src/resolver.rs
```diff
@@ -11,8 +11,85 @@ use serde_json::Value;
...
+/// struct MyCustomResolver;
+///
+/// impl SchemaResolver for MyCustomResolver {
+///     fn resolve(&self, root_schema: &Value, url: &Url) -> Result<Arc<Value>, SchemaResolverError> {
+///         match url.scheme() {
+///             "json-schema" => {
+///                 Err(CustomError::new("cannot resolve schema without root schema ID"))
+///             },
+///             "http" | "http" => {
```
Stranger6667, line 36:
```suggestion
///             "http" | "https" => {
```

jsonschema/src/resolver.rs
```diff
@@ -55,25 +144,19 @@ impl Resolver {
...
-                                let arcdoc = Arc::new(document);
-                                self.store
-                                    .write()
-                                    .insert(url_str.to_string(), arcdoc.clone());
-                                Ok(arcdoc)
-                            }
-                            #[cfg(not(any(feature = "reqwest", test)))]
-                            panic!("trying to resolve an http(s), but reqwest support has not been included");
+                    None => {
+                        if let Some(s) = self.store.read().get(url_str) {
```
Stranger6667, line 148: Hm, it should not be necessary, as `store` is checked a few lines above

Stranger6667 (changes requested):
Amazing! Thank you so much @tamasfe for putting this PR with such a great level of details :) 
I added a few minor comments, and here are a few more general things:
- I am good with adding `anyhow`, it is indeed quite convenient :)
- Re: benchmarks - I don't think there will be notable changes + caching will reduce it even more
- A changelog entry would be nice
- There are failing tests in Python bindings, but it should be solved by adding new `resolve-*` features

Otherwise, looks great to me!

jsonschema/src/error.rs
tamasfe: I don't think so, as even if `resolve-http` and `resolve-file` are not selected, the user of the library can still provide their own resolver.

If it was put behind `cfg`, `SchemaResolver` and all related items should be behind `cfg` as well.

Also, in the previous version, a panic happened if the feature was not enabled, but now it's an error instead. While technically not a validation error, I'd still generally prefer this behaviour where possible as accidental panics can bring down entire applications.

Stranger6667: Aha, makes sense :)

jsonschema/src/resolver.rs
tamasfe: Indeed, I removed the earlier check instead as it requires a lock, not that it would matter too much.

jsonschema/Cargo.toml
tamasfe: Actually I lied about the feature changes not breaking things, they might still be breaking for users that use `default-features = false` and `features = ["reqwest"]`, although it's not something incredibly hard to fix.

I added a compile error so that people who only have the `reqwest` feature enabled will be notified of the change instead of silently breaking references.

tamasfe:
Thanks @Stranger6667!

I've also added the original schema reference string to be passed to the resolver so that relative references are always preserved.

The tests should pass now as well. I also started building on `jsonschema` along with this PR in my project with great results so far.

Stranger6667 (approved):

Stranger6667:
Thank you @tamasfe for your contribution! :) Great PR!

> I also started building on jsonschema along with this PR in my project with great results so far.

Glad to hear about it!

Stranger6667:
Could you, please, rebase the branch?

Stranger6667:
Strange, that after rebase the UI still showed some conflicts, so I squashed it :) Will make a new release soon

tamasfe:
Awesome, thanks!

Github also showed that my branch was based on master HEAD, so I'm not entirely sure why the rebase was needed (but it was). It was probably due to making edits on the GUI here.

Sorry about the awful graph, I don't think I'll ever figure git out in this lifetime.

Issue #343, opened 2022-02-24
HTTP Resolver Not Working?


rawkode:
Hi,

Sorry, I've been trying to work this out myself, but I'm not sure what's going wrong.

```
Validation error: failed to resolve https://schema.rawkode.dev/organization: error sending request for url (https://schema.rawkode.dev/organization): error trying to connect: invalid URL, scheme is not http
Instance path:
Validation error: failed to resolve https://schema.rawkode.dev/person: error sending request for url (https://schema.rawkode.dev/person): error trying to connect: invalid URL, scheme is not http
Instance path:
```

These URLs are definitely OK and I'm not sure what I've done wrong.

I've tried using `http` instead of `https` in my schema, but the problem is the same.

Any help appreciated.

Stranger6667:
Hi!

Thanks for reporting! Could you, please, share the schema so I can reproduce the issue?

Looks strange to me - does it work with `0.14.0`? There were some changes with reference resolving

rawkode:
Sure, the schema is:

```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
        "people": {
            "type": "array",
            "items": {
                "$ref": "https://schema.rawkode.dev/person"
            },
            "links": [
                {
                    "rel": "collection",
                    "href": "#/organizations/{id}",
                    "templateRequired": [
                        "id"
                    ]
                }
            ]
        },
        "organizations": {
            "type": "array",
            "items": {
                "$ref": "https://schema.rawkode.dev/organization"
            }
        }
    }
}
```

I'll test with 0.14 and report back.

rawkode:
Still a problem with `0.14`. I hope this doesn't mean its my fault 😅 

Let me know if you need anything else, @Stranger6667

Stranger6667:
Thanks for sharing!

At 778b424 I cannot reproduce it locally :(

Tried this instance:

```json
{
    "people": [
        {"name": "x", "forename": "y", "surname":  "z"}
    ],
    "organizations": [
        {"name": "a"}
    ]
}
```

And this command (I suppressed warnings coming from not enabled newer drafts support):

```
⇒  RUSTFLAGS=-Awarnings cargo r -- schema.json -i instance.json
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/jsonschema schema.json -i instance.json`
instance.json - VALID
```

[...]

PR #344, opened 2022-02-25
test: Add a test case for #343

Stranger6667:


Stranger6667:
The underlying issue should be resolved by #351

PR #345, opened 2022-02-28
fix!: specify precise dependency versions in `Cargo.toml` and upgrade all versions

fujiapple852:
As of the latest master (778b424b3de65f2ed6df64a2558bde7341c0094b) the library does not compile with cargo `minimal-versions` for the default features:

```shell
cargo +nightly update -Z minimal-versions && cargo +nightly check
```

This PR:
- Specifies precise dependency versions in `Cargo.toml` (as discussed [here](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277))
- Upgrades all dependancies to the latest compatible versions such that the test suite passes with `--all-features`

Note that I was not able to determine the minimal version updates to allow the library to build with `minimal-versions` and so I took the sledge-hammer approach and upgraded everything.

And finally, thank you for proving this library to the community!

### DIFF
--- a/jsonschema/Cargo.toml
+++ b/jsonschema/Cargo.toml
@@ -25,37 +25,37 @@ resolve-http = ["reqwest"]
 resolve-file = []
 
 [dependencies]
-ahash = { version = "0.7", features = ["serde"] }
-anyhow = "1"
-base64 = ">= 0.2"
-bytecount = { version = "0.6", features = ["runtime-dispatch-simd"] }
-fancy-regex = "^0.7.1"
-fraction = { version = "0.9", default-features = false, features = ["with-bigint"] }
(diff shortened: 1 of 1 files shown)


fujiapple852:
@Stranger6667 I'm not clear on the cause of the CI failures for the windows tests, do you have any insight on this?  I can see a similar failure for an unrelated CI run so perhaps the failure is unrelated to this specific PR.

Stranger6667:
Hm, not sure - it happened somewhere before as well (not sure which bullied it was) I will get back to this in a few days

Stranger6667:
It took longer than I anticipated :) Thank you for making this PR, I'll issue a patch release shortly

PR #354, opened 2022-04-05
feat: custom keywords

antouhou:
Hi @Stranger6667! Thank you very much for creating this awesome library! 

I've written some code that adds ajv-ish way to define custom keywords. Would you be considering adding something like this, and if so, can you have any comments on this PR? Thank you!

### DIFF
--- a/jsonschema/src/compilation/options.rs
+++ b/jsonschema/src/compilation/options.rs
@@ -10,6 +10,8 @@ use crate::{
 };
 use ahash::AHashMap;
 use std::{fmt, sync::Arc};
+use serde_json::Value;
+use valico::json_schema::validators::Validator;
 
 const EXPECT_MESSAGE: &str = "Valid meta-schema!";
 
@@ -221,7 +223,7 @@ pub struct CompilationOptions {
     formats: AHashMap<&'static str, fn(&str) -> bool>,
     validate_formats: Option<bool>,
     validate_schema: bool,
-    // custom_keywords: Option(HashMap<String, Validator>
+    custom_keywords: AHashMap<String, KeywordDefinition>,
 }
 
 impl Default for CompilationOptions {
@@ -235,6 +237,7 @@ impl Default for CompilationOptions {
             store: AHashMap::default(),
             formats: AHashMap::default(),
             validate_formats: None,
+            custom_keywords: AHashMap::default(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Stranger6667:
Hi @antouhou! 

Thank you so much for this PR! I am definitely interested in having support for custom keywords :) I'll take a look at it next weekend

Happy to hear that you find this library useful! :)

jsonschema/src/keywords/custom_keyword_schema.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::keywords::CompilationResult;
+use crate::paths::{InstancePath, JSONPointer, PathChunk};
+use crate::validator::Validate;
+use crate::{ErrorIterator, JSONSchema, ValidationError};
+use serde_json::{Map, Value};
+use std::fmt::{Display, Formatter};
+
+pub(crate) struct CustomKeywordSchemaValidator {
+    schema_path: JSONPointer,
+    keyword_schema: Option<Value>,
```
Stranger6667, line 13: What would be the use case for `None`? 

As far as I can tell, if one wants to create a custom keyword, then there is always some value there, even if it is a marker-like `null` value. I don't think that there is a place for niche optimization for `Option<Value>`, so there should be no memory win there anyway. 

Therefore I am thinking that the plain `Value` could be the way to go here. What do you think?

jsonschema/src/keywords/custom_keyword_schema.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[inline]
+    pub(crate) fn compile(
+        _: &Value,
+        schema_path: JSONPointer,
+        keyword_schema: Value,
+    ) -> CompilationResult {
+        let mut validator = CustomKeywordSchemaValidator {
+            schema_path,
+            keyword_schema: Some(keyword_schema),
+            json_schema: None,
```
Stranger6667, line 27: The general direction in other keywords is to avoid perf costs at runtime - if it could be detected at the compilation phase, then it is better to avoid it altogether.

If there is nothing in the child validator, then the parent one should not be compiled at all. For this reason, all the `compile` functions return `Option<CompilationResult<'a>>`.

I'd suggest rearranging the implementation a bit, so the `json_schema` field has the `JSONSchema` type and always has something to validate.

jsonschema/src/compilation/options.rs
```diff
@@ -568,7 +571,25 @@ impl CompilationOptions {
...
+
+    pub(crate) fn custom_keyword_definition(&self, keyword: &str) -> Option<&KeywordDefinition>
+    {
+        self.custom_keywords.get(keyword)
+    }
 }
+
+#[derive(Clone)]
+pub enum KeywordDefinition {
+    Schema(Value)
```
Stranger6667, line 590: What are the possible other variants here?

Stranger6667 (commented):
Hi @antouhou 

Thank you again for the PR and sorry for not getting back to you earlier.

I definitely like this PR and mostly have some minor nitpicking on naming, a couple of questions for the possible use cases, and some context about the common implementation approach in this library.

Great work! :) Let me know what you think about those comments, and I'd be happy to collaborate on this PR further!

PR #355, opened 2022-04-10
Allow URL schema resolving with rustls

jqnatividad:
Hi @Stranger6667,
I'm also using jsonschema-rs as a library, and needed to use `rustls` with URL schemas, as cross no longer bundles openssl which `native-tls` requires. 

As per the README, I explicitly enabled `reqwest` and `rustls` and disabled `native-tls` via `default-features = false` in my Cargo.toml file but I still got the compilation error.

```
jsonschema = { version = "0.15.1", features = [
    "resolve-file",
    "reqwest",
    "rustls"
], default-features = false }
```

```
Compiling jsonschema v0.15.1
error: the `reqwest` feature does not enable HTTP schema resolving anymore, use the `resolve-http` feature instead
  --> C:\Users\jsmith\.cargo\registry\src\github.com-1ecc6299db9ec823\jsonschema-0.15.1\src\resolver.rs:81:21
   |
81 | ...   compile_error!("the `reqwest` feature does not enable HTTP schema resolving anymore, use the `resolve-http` feature instead");
```

I had to tweak the conditional compilation directives to get it to work.

Fixes #353

PR #360, opened 2022-04-21
fix: Various compilation issues

### LINKED ISSUE #356: Compilation fails with no default features
`0.15.1` and `0.15.2` introduced feature changes/additions I don't necessarily agree with so I don't know what kind of PR to submit for this.

Having both a `resolve-http` and `reqwest` features that slightly differ in semantics makes no sense, since `resolve-http` includes the `reqwest` feature, it alone should be enough to enable the HTTP(S) schema resolution feature.

Also I believe exposing optional features of dependencies is just a way to pollute the features of this library and also leads to issues like this one. #343 could have been solved simply by suggesting the addition of `reqwest = { version = "*", features = [ "rustls-tls" ] }`, the current solution raises the question why only `native-tls` and `rustls` are exposed while `reqwest` has a handful more of TLS features and combinations.

### LINKED ISSUE #358: Allow compiling only with resolve-file feature
If jsonschema is included using the following file:

```
jsonschema = { version = "0.15.2", default-features = false, features = ["resolve-file"] }
```

builds will fail at https://github.com/Stranger6667/jsonschema-rs/blob/master/jsonschema/src/resolver.rs#L85 because of the compile_error. I'm in a situation where I *only* want to load schemas off local storage and do not need http resolution access.

The same check happens at line 102 (https://github.com/Stranger6667/jsonschema-rs/blob/master/jsonschema/src/resolver.rs#L102) in order to throw an error, but we'll never reach that because of the compile error so we can't even build in the first place.

### LINKED ISSUE #359: reqwest is not optional with version 0.15.2 ?
I am trying to compile without reqwest(which has lot of dependency) and this the following code breaks the compilation.

https://github.com/Stranger6667/jsonschema-rs/blob/e2d09aa932145db970556eb1fe1ed6edb824a270/jsonschema/src/resolver.rs#L78-L90

`rust-v0.15.1` tag does not have this problem.

Stranger6667:
Resolves #356
Resolves #358
Resolves #359

@tamasfe @jqnatividad @samueldotj @garypen @qdot @rawkode

Sorry for tagging, but as you opened issues about compilation problems or features not working properly, I assume, this change might affect you. I'd like to let you know about this change upfront, so you can try it out and share your concerns. 

It effectively reverts #351 and #355, so if one wants to enable a specific TLS backend, then it requires an explicit `reqwest` feature:

```toml
reqwest = { version = "*", features = [ "rustls-tls" ] }
```

I am also open to submitting PRs to your open repos when the new `jsonschema` version is released, let me know if I can help there.

As a separate measure, I'd also work on including [cargo-hack](https://github.com/taiki-e/cargo-hack) into the pipeline to avoid compilation errors due to feature config in the future.

Once again, sorry for the complication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Library compilation with no default features. [#356](https://github.com/Stranger6667/jsonschema-rs/issues/356)
+- Compilation with `resolve-file` only. [#358](https://github.com/Stranger6667/jsonschema-rs/issues/358)
+
+### Changed
+
+- **BREAKING**: Revert changes from [#353](https://github.com/Stranger6667/jsonschema-rs/issues/353) and [#343](https://github.com/Stranger6667/jsonschema-rs/issues/343), as they caused compilation issues.
+
 ## [0.15.2] - 2022-04-10
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -23,9 +23,6 @@ Partially supported drafts (some keywords are not implemented):
 jsonschema = "0.15"
 ```
 
-By default `jsonschema` resolves remote references via HTTP by using `reqwest` with `native-tls`.
-If you'd like to use `rustls`, you have to explicitly enable `reqwest` and `rustls` features and disable `native-tls` via `default-features = false` in your `Cargo.toml` file.
-
 To validate documents against some schema and get validation errors (if any):
 
 ```rust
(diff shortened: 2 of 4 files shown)


tamasfe:
I was only mildly affected and this should not require any actions from me after a release. I just wanted to thank you again for the great work on this library, and the prompt fix as well!

qdot:
Yeah, no need to submit anything to my repos either, this only occurred on a experimental android branch I was building. Thanks for the library, it's great! :)

PR #375, opened 2022-05-27
Improve resolving

Stranger6667:


Stranger6667:
This implementation is actually slower because the exploration stack lives on the heap. Though, it might be better to use the program stack instead - it might be faster (as unrolling multiple levels manually actually gives a lot in performance)

PR #384, opened 2022-09-15
Replace `structopt` with `clap`

Expyron:


Stranger6667:
What is the reason for this change?

Expyron:
Structopt is not maintained anymore: https://github.com/TeXitoi/structopt/issues/525
Its author recommends moving to clap v3.

Julian:
As is I think this also is triggering GitHub security warnings for users of `jsonschema-rs` -- e.g. on bowtie, due to the transitive dependency on `ansi_term`, I see:

Screen Shot 2022-09-17 at 14 47 01

Stranger6667:
Sorry for the delay! Merged :) I'll make a new release soon

PR #387, opened 2022-10-18
fix(python): Fix SystemError when non-str dict key provided #386

### LINKED ISSUE #386: SystemError: <method 'validate' of 'jsonschema_rs.JSONSchema' objects> returned a result with an exception set
### Prerequisites
* jsonschema-rs version:  0.16.0
* Python version: 3.10.2
* Operating System: macos

### Description
System error when trying to check dict with numeric keys.

Example:
```python
from jsonschema_rs import JSONSchema

schema = JSONSchema({'type': 'object', 'additionalProperties': {'type': 'string'}})

schema.validate({123: "123"})
```

Result:
```bash
TypeError: bad argument type for built-in operation

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "...../test.py", line 5, in <module>
    schema.validate({123: "123"})
SystemError: <method 'validate' of 'jsonschema_rs.JSONSchema' objects> returned a result with an exception set

Process finished with exit code 1
```

### Expected behavior

Validation error or pass validation

ermakov-oleg:
Fix #386

### DIFF
--- a/bindings/python/src/ser.rs
+++ b/bindings/python/src/ser.rs
@@ -18,7 +18,7 @@ use std::ffi::CStr;
 
 pub const RECURSION_LIMIT: u8 = 255;
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub enum ObjectType {
     Str,
     Int,
@@ -140,6 +140,15 @@ impl Serialize for SerializePyObject {
                     for _ in 0..length {
                         unsafe {
                             pyo3::ffi::PyDict_Next(self.object, &mut pos, &mut key, &mut value);
+                        }
+                        match get_object_type_from_object(key) {
+                            ObjectType::Str => {}
+                            object_type => {
+                                return Err(ser::Error::custom(format!(
+                                    "Supported only str key type. Provided type '{:?}'",
+                                    object_type
+                                )))
+                            }
                         }
                         let uni = unsafe { string::read_utf8_from_str(key, &mut str_size) };
                         let slice = unsafe {
--- a/bindings/python/src/ser.rs
+++ b/bindings/python/src/ser.rs
@@ -74,6 +74,23 @@ fn get_object_type_from_object(object: *mut pyo3::ffi::PyObject) -> ObjectType {
     }
 }
 
+fn get_type_name(object_type: *mut pyo3::ffi::PyTypeObject) -> String {
+    let type_name = unsafe { CStr::from_ptr((*object_type).tp_name).to_string_lossy() };
+    type_name.to_string()
+}
+
+#[inline]
+fn check_type_is_str<E: ser::Error>(object: *mut pyo3::ffi::PyObject) -> Result<(), E> {
+    let object_type = unsafe { Py_TYPE(object) };
+    if object_type != unsafe { types::STR_TYPE } {
+        return Err(ser::Error::custom(format!(
+            "Dict key must be str. Got '{}'",
+            get_type_name(object_type)
+        )));
+    }
+    Ok(())
+}
+
 #[inline]
 pub fn get_object_type(object_type: *mut pyo3::ffi::PyTypeObject) -> ObjectType {
     if object_type == unsafe { types::STR_TYPE } {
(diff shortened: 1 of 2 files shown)


bindings/python/src/ser.rs
```diff
@@ -141,6 +141,15 @@ impl Serialize for SerializePyObject {
                         unsafe {
                             pyo3::ffi::PyDict_Next(self.object, &mut pos, &mut key, &mut value);
                         }
+                        match get_object_type_from_object(key) {
+                            ObjectType::Str => {}
+                            object_type => {
+                                return Err(ser::Error::custom(format!(
+                                    "Supported only str key type. Provided type '{:?}'",
```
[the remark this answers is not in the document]
ermakov-oleg: Yes, thank you, that would be better.

bindings/python/src/ser.rs
```diff
@@ -74,6 +74,23 @@ fn get_object_type_from_object(object: *mut pyo3::ffi::PyObject) -> ObjectType {
...
+    type_name.to_string()
+}
+
+#[inline]
+fn check_type_is_str<E: ser::Error>(object: *mut pyo3::ffi::PyObject) -> Result<(), E> {
+    let object_type = unsafe { Py_TYPE(object) };
+    if object_type != unsafe { types::STR_TYPE } {
+        return Err(ser::Error::custom(format!(
+            "Dict key must be str. Got '{}'",
+            get_type_name(object_type)
```
Stranger6667, line 88: Nit: I think in this case it should be possible to avoid one `String` allocation inside `get_type_name`. See another comment

bindings/python/src/ser.rs
```diff
@@ -74,6 +74,23 @@ fn get_object_type_from_object(object: *mut pyo3::ffi::PyObject) -> ObjectType {
     }
 }
 
+fn get_type_name(object_type: *mut pyo3::ffi::PyTypeObject) -> String {
+    let type_name = unsafe { CStr::from_ptr((*object_type).tp_name).to_string_lossy() };
```
Stranger6667, line 78: `type_name` is `Cow<str>`, and likely its borrowed variant (as the type name is likely to be a valid UTF-8 string). Allocating a new `String` on the following line is not needed when it is passed to `format!` inside `check_type_is_str` later on.

This code path is quite rare, and it is just saves a single allocation, so feel free to skip this.

Stranger6667 (approved):
Cool! There is a minor perf nitpick, but I am happy to merge it as is

bindings/python/src/ser.rs
ermakov-oleg: In the case of `Cow` it is necessary to specify a `'static` lifetime parameter, otherwise the compiler gets a error. 
Which does not seem very true, since the type can be deleted.

I'm not very good at rust yet, maybe specifying it as static would be correct.

Stranger6667: Indeed, the lifetime is bound to the Python interpreter memory flow, which we can assume to be `'static` at this point (which is not 100% true, as you mentioned). CPython guarantees that GIL is held when this extension gets the control - no other thread can remove the type from Python's heap. Here we also don't spin anything else that can affect the type, so, given these circumstances I'd say it should be safe to use `Cow<'static, str>` in this case. The underlying slice will be accessed only a little longer (`format!` will create an owned string just after this). Technically, we already hold `Cow<'static, str>` here in `type_name`, such a change won't change the implementation conceptually (it is still "type pointer" -> "CStr" -> "Cow" -> "String")

Stranger6667: Or, better - it should be the `'p` lifetime, which we got from the original `&PyAny` "owned reference", that is GIL-bound.

ermakov-oleg: Okay, thanks for such a detailed answer.
In that case, yes, it looks like we won't break any invariants, and the reference to the type name will only be stored until the `format!` execution.

I can fix that in a new pull request.

ermakov-oleg: Didn't see that you had already fixed it)
Thank you!

Stranger6667: Sorry, I forgot to let you know about it :)

Stranger6667: I'll make a new release soon

PR #392, opened 2022-10-26
fix(python): Add .pyi "stub" file for type checkers support #391

### LINKED ISSUE #391: [Python] Add .pyi "stub" file for type checkers support
I can try to do it in my free time

ermakov-oleg:
Closes #391


ermakov-oleg:
I'm not sure if I need to add type checks to CI, since the result of the `mypy` run is mostly trash (missing a type annotations for tests). 
The `pyright` situation is slightly better, for example it can be run for directory with tests (`tests-py`)

bindings/python/jsonschema_rs.pyi
```diff
@@ -0,0 +1,69 @@
...
+    def validate(self, instance: Any) -> None:
+        pass
+
+    def iter_errors(self, instance: Any) -> Iterator[ValidationError]:
+        pass
+
+
+class ValidationError(Exception):
+    message: str
+    schema_path: list[Union[str, int]]
```
Stranger6667, line 63: Should we use ‘List’ instead to support users running type checkers on older Python versions?

ermakov-oleg: As far as I remember, it is possible to use the last features, because the interpreter does not look at these files

ermakov-oleg: https://typing.readthedocs.io/en/latest/source/stubs.html#syntax

ermakov-oleg: It turns out we can even go to the new unions syntax

Stranger6667: Awesome! Thank you for clarification

bindings/python/jsonschema_rs.pyi
```diff
@@ -0,0 +1,69 @@
+from typing import Any, TypeVar
+from collections.abc import Iterator
```
Stranger6667, line 2: Should it be from `typing` as well?

bindings/python/jsonschema_rs.pyi
```diff
@@ -0,0 +1,69 @@
...
+        pass
+
+    def validate(self, instance: Any) -> None:
+        pass
+
+    def iter_errors(self, instance: Any) -> Iterator[ValidationError]:
+        pass
+
+
+class ValidationError(Exception):
```
Stranger6667, line 61: It is actually [inherited](https://github.com/Stranger6667/jsonschema-rs/blob/40498fe12023a0f5825fefa54f955a355fe132e3/bindings/python/src/lib.rs#L39) from `ValueError`

bindings/python/jsonschema_rs.pyi
ermakov-oleg: It's already deprecated https://docs.python.org/3/library/typing.html#typing.Iterator

Stranger6667: Thanks!

Stranger6667:
Just to confirm - should it be distributed with the package itself?

ermakov-oleg:
> Just to confirm - should it be distributed with the package itself?

Hmm, this does seem to be a problem, since locally I built with maturin and it automatically adds `py.typed` to the package and transfers the type annotations to `__init__.pyi`

ermakov-oleg:
Yes, maturin and github action makes it much easier to build by building all the necessary wheels at once. Right now they are very lacking for macos arm.

ermakov-oleg:
I accidentally pressed sync on my fork and gh closed PR...

Stranger6667 (approved):
Looks good to me! Thank you :)

Issue #379, opened 2022-06-17
Custom Validators


tamasfe:
I got a feature request in a project that uses the library to validate file paths in a document. It would make sense to _simply_ add a validator that would be able to do this, but it's currently not possible. The closest I can get is format validators but they lack any context.

Would it be desirable to expose a way to hook into the compilation process and add entirely custom validators?

This might _not_ be a good idea because:

- it has nothing to do with the JSON schema spec
- complexity and larger API surface

However it would be very convenient for my use-case.

If this is desirable I imagine in order to do this we'd need to:

- expose the `Validate` trait or something similar
- expose a way to hook into the compilation process and create validators
- add a custom error type that could be pretty much anything (probably `Any + Display`)
- add a way to pass contextual information to validators to avoid recompilations (optional, I have no idea how much complexity this would add)

samgqroberts:
I'd just like to add on to this my own use case, as additional information about how people are using the project and working around this particular issue (+ #245 ).

I use the Python binding, but I have a custom string format called "currency" which means I can't use this library for all of my validation. The "currency" format requires that a string has some amount of digit characters, followed by a period, followed by exactly two digit characters. It is meant to work around the fact that JSON does not have a fixed-point decimal type, so currency figures are stored in strings to maintain precision / avoid floating point nonsense.

I have a wrapper function around the validation that first checks whether the provided schema contains (recursively) any "currency" format:

```python
def detect_currency_format(schema: Any) -> Any:
    if isinstance(schema, list):
        return any([detect_currency_format(x) for x in schema])
    if isinstance(schema, dict):
        if schema.get("format") == "currency":
            return True
        return any([detect_currency_format(v) for v in schema.values()])
    return False
```

If the schema does contain the format, I know I need to use another library to validate that entire schema. The other library I use is [jsonschema](https://python-jsonschema.readthedocs.io/en/stable/). There I can add custom formats, which I do for "currency", and provide a lambda for the validation which just calls out to a regex.

Furthermore, the other library allows for overriding keyword validation in general. I use that to override default behavior for the numeric modifiers "multipleOf", "minimum", "maximum", "exclusiveMinimum", and "exclusiveMaximum" so that they apply to currency strings as if they were any other numeric type.

[...]

Issue #342, opened 2022-02-14
Custom Keywords


sacvalleytech:
I just ran across this library and had a look at the docs... I didn't see any examples of how to extend with a custom keyword. Is this possible?

antouhou:
I've created a PR that adds the ability to define a custom keyword schema. I think if that's something you would be interested in, you can give it a try: https://github.com/Stranger6667/jsonschema-rs/pull/354

Stranger6667:
Implemented in aa94a4b24aee8b4c7d6074c6aaf413176672528b

PR #394, opened 2022-11-02
feat: Custom keyword validation

bnjt:
Add support for user defined custom keyword validation. The user provides and registers custom validator functions when configuring a JSONSchema.

Custom keyword validators may be used when the user wants to enforce constraints that can't, or can't easily, be expressed in JSON schema.

This PR addresses at least some of the features requested in #379 by @tamasfe as well as my own. The same pros and cons described in #379 apply. The PR also follows the structure of #354, but of course uses validation functions as opposed to sub-schema for validation.

We could collapse the two user implementable validation functions into a single trait, mirroring the internal `Validate` trait. Implementations would have to sit behind an `Arc`.

Example usage
```rust
use jsonschema::{CustomKeywordDefinition, ErrorIterator, JSONSchema, paths::JSONPointer};
use serde_json::{json, Value};
use std::sync::Arc;

fn validate(
    instance: &Value,
    instance_path: JSONPointer,
    schema: Arc<Value>,
    schema_path: JSONPointer,
) -> ErrorIterator {
    // ... validate instance ... 
    Box::new(None.into_iter())
}
fn is_valid(instance: &Value, schema: &Value) -> bool {
    // ... determine if instance is valid ... 
    return true;
}
let definition = CustomKeywordDefinition::Validator { validate, is_valid };
assert!(JSONSchema::options()
    .with_custom_keyword("my-type", definition)
    .compile(&json!({ "my-type": "my-schema"}))
    .expect("A valid schema")
    .is_valid(&json!({ "a": "b"})));
```

### DIFF
--- a/jsonschema/src/compilation/options.rs
+++ b/jsonschema/src/compilation/options.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 use ahash::AHashMap;
 use std::{
-    fmt::{self, Display},
+    fmt::{self},
     sync::Arc,
 };
 
--- a/jsonschema/src/compilation/options.rs
+++ b/jsonschema/src/compilation/options.rs
@@ -5,11 +5,15 @@ use crate::{
         DEFAULT_CONTENT_ENCODING_CHECKS_AND_CONVERTERS,
     },
     content_media_type::{ContentMediaTypeCheckType, DEFAULT_CONTENT_MEDIA_TYPE_CHECKS},
+    keywords::custom_keyword::{CustomIsValidFn, CustomValidateFn},
     resolver::{DefaultResolver, Resolver, SchemaResolver},
     schemas, ValidationError,
 };
 use ahash::AHashMap;
-use std::{fmt, sync::Arc};
+use std::{
+    fmt::{self, Display},
+    sync::Arc,
+};
 
 const EXPECT_MESSAGE: &str = "Valid meta-schema!";
 
@@ -222,6 +226,7 @@ pub struct CompilationOptions {
     validate_formats: Option<bool>,
     validate_schema: bool,
     ignore_unknown_formats: bool,
+    custom_keywords: AHashMap<String, CustomKeywordDefinition>,
 }
 
 impl Default for CompilationOptions {
@@ -236,6 +241,7 @@ impl Default for CompilationOptions {
             formats: AHashMap::default(),
             validate_formats: None,
             ignore_unknown_formats: true,
+            custom_keywords: AHashMap::default(),
         }
     }
 }
@@ -584,6 +590,50 @@ impl CompilationOptions {
     pub(crate) const fn are_unknown_formats_ignored(&self) -> bool {
         self.ignore_unknown_formats
     }
+
+    /// Register a custom keyword definition.
+    ///
+    /// Examples
+    ///
+    /// ```rust
+    /// # use jsonschema::{CustomKeywordDefinition, ErrorIterator, JSONSchema, paths::JSONPointer};
+    /// # use serde_json::{json, Value};
+    /// # use std::sync::Arc;
+    ///
+    /// fn validate(
+    ///     instance: &Value,
+    ///     instance_path: JSONPointer,
+    ///     schema: Arc<Value>,
+    ///     schema_path: JSONPointer,
+    /// ) -> ErrorIterator {
+    ///     // ... validate instance ...
+    ///     Box::new(None.into_iter())
+    /// }
+    /// fn is_valid(instance: &Value, schema: &Value) -> bool {
+    ///     // ... determine if instance is valid ...
+    ///     return true;
+    /// }
+    /// let definition = CustomKeywordDefinition::Validator { validate, is_valid };
+    /// assert!(JSONSchema::options()
+    ///     .with_custom_keyword("my-type", definition)
+    ///     .compile(&json!({ "my-type": "my-schema"}))
+    ///     .expect("A valid schema")
+    ///     .is_valid(&json!({ "a": "b"})));
+    /// ```
+    pub fn with_custom_keyword<T>(mut self, keyword: T, definition: CustomKeywordDefinition) -> Self
+    where
+        T: Into<String>,
+    {
+        self.custom_keywords.insert(keyword.into(), definition);
+        self
+    }
+
+    pub(crate) fn get_custom_keyword_definition(
+        &self,
+        keyword: &str,
+    ) -> Option<&CustomKeywordDefinition> {
+        self.custom_keywords.get(keyword)
+    }
 }
 // format name & a pointer to a check function
 type FormatKV<'a> = Option<(&'a &'static str, &'a fn(&str) -> bool)>;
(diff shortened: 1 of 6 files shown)


bnjt:
Hi @Stranger6667, Do you think this feature PR is aligned with the libraries goals? I can appreciate it might not add value to enough of users to make it the addition and maintenance of the extra code worth it. However, if it is a use case you're interested supporting, I'd be happy to work on any improvements necessary to make it a better fit.

Stranger6667:
Hey! Thank you for pinging me! I’ll take a look this week

Stranger6667:
Thank you so much and sorry for my late reply!

I like the simplicity of providing two functions & the API, however, having a trait allows the end user to carry some extra data within a custom struct that implements such a trait. It might make this feature more flexible and cover more use cases. What do you think?

Stranger6667:
@bnjt

I like the implementation very much! I have a few bikeshedding ideas regarding naming, but I'd like to discuss them later on. The main two points I'd like to discuss are:

1. Using trait instead of two functions to allow the users to carry some extra data in a struct. What user-facing API could it have? would it be convenient to use?
2. Extending the `CustomKeywordDefinition` enum in the future. What use cases could we think of in order to avoid possible backward-incompatible changes? I guess it is an alternative to having a trait

I don't have many specific thoughts about these topics yet, but I'd like to hear your opinion :)

cc @tamasfe As you created #379 I'd love to hear your thoughts as well :) 

Tagging @antouhou as the author of #354 and @hadrien-toma & @Arcahub as you reacted on that PR. It would be amazing if you could share your use cases & whether this PR would cover them.

Stranger6667:
Also cc-in @ryderjgillen as the author of #342 (and @manuschillerdev as you reacted there). Your input is welcome too! :)

tamasfe:
> As you created https://github.com/Stranger6667/jsonschema-rs/issues/379 I'd love to hear your thoughts as well :)

Thanks, this is along the lines of what I described more or less, if it was up to me I'd take it a step further and not try to restrict it to custom keywords, I'd just have a factory exposed to users that looks something like `Fn(schema, path) -> Option<Box<dyn Validator>>` that hooks into the compilation process and works for every schema regardless of keywords.

And `Validator` is pretty much the trait what already exists internally, one more thing I'd add is a catch-all error type for user-provided validator errors.

bnjt:
Hi @Stranger6667, thank you for the response. Apologies for being late on my side. 

#### Re point 1 to use a trait
> Using trait instead of two functions to allow the users to carry some extra data in a struct. 

I agree. I don't recall, but I _think_ I might have gone that route and then backed out due to needing a lifetime. But a trait is a cleaner approach and I'm sure whatever issue I ran into is solvable.

> What user-facing API could it have? 

Much like @tamasfe I was thinking a public version of [Validate](https://github.com/Stranger6667/jsonschema-rs/blob/f44f195dc1eb2f400b776b928a320d6454a2ba00/jsonschema/src/validator.rs#L25) but narrowed down to just
- `validate(instance, path) -> ErrorIterrator` and
- `is_valid(instance) -> bool 

> would it be convenient to use?

IMHO that's a friendly API. We could also provide a trait default implementation for `is_valid()`.

Simple implementations with immutable state are simple. Implementers who want mutable state, would have expend some effort to provide their own interior mutability. But that seems reasonable to me.

####  Re. point 2 Extending the CustomKeywordDefinition

> Extending the CustomKeywordDefinition enum in the future. What use cases could we think of in order to avoid possible backward-incompatible changes? I guess it is an alternative to having a trait

I don't have any immediate suggestions, but will give this some thought along with @tamasfe's suggestion to generalize.

jsonschema/src/compilation/options.rs
```diff
@@ -601,6 +651,34 @@ impl fmt::Debug for CompilationOptions {
...
+    ///
+    /// A custom keyword validator may be used when a validation that cannot, or
+    /// cannot be be easily or efficiently expressed in JSON schema.
+    ///
+    /// The custom validation is applied in addition to the JSON schema validation.
+    Validator {
+        /// Validate an instance returning any and all detected validation errors
+        validate: CustomValidateFn,
+        /// Determine if an instance is valid
+        is_valid: CustomIsValidFn,
```
samgqroberts, line 670: Since the distinction between `validate` and `is_valid` only exists to improve the performance of `is_valid`, we could allow the user to omit this by making this an `Option<CustomIsValidFn>`, and have the `is_valid()` implementation of the `Validate` trait have a default case if this is missing that calls `validate` and returns if any errors are present. This would allow the user to make the tradeoff of losing out on `is_valid()` perf a bit but avoiding duplication of the validation logic.

jsonschema/src/compilation/options.rs
```diff
@@ -584,6 +590,50 @@ impl CompilationOptions {
...
+    ///     return true;
+    /// }
+    /// let definition = CustomKeywordDefinition::Validator { validate, is_valid };
+    /// assert!(JSONSchema::options()
+    ///     .with_custom_keyword("my-type", definition)
+    ///     .compile(&json!({ "my-type": "my-schema"}))
+    ///     .expect("A valid schema")
+    ///     .is_valid(&json!({ "a": "b"})));
+    /// ```
+    pub fn with_custom_keyword<T>(mut self, keyword: T, definition: CustomKeywordDefinition) -> Self
```
samgqroberts, line 623: I believe this does not need to consume `self`, so maybe make it take in and return `&mut self` just like `with_format()` above? This would allow it to chain method calls more easily.

jsonschema/src/compilation/options.rs
samgqroberts: Ah I see @Stranger6667 noted this in his comment just above, wrt the trait default implementation

samgqroberts:
Hi @Stranger6667 and @bnjt. Thank you for all your work here! I took a look myself and wanted to make sure this feature would work for my use case, which I describe over here https://github.com/Stranger6667/jsonschema-rs/issues/379#issuecomment-1643959341.

It was 90% of the way there, so I tried to take it the remaining 10% with a PR against this PR: https://github.com/bnjt/jsonschema-rs/pull/1

Please let me know what you think!

platoscave:
Is this feature available to the public yet?
 Which version?

samgqroberts:
@platoscave as far as I know this was never merged to the project. I made a fork a while ago that merged @bnjt's work plus an additional piece I needed for my use case I mention in an earlier comment. The last 3 commits of the fork's main branch show that work https://github.com/samgqroberts/jsonschema-rs/commits/master/. For what it's worth I've been using that version of jsonschema-rs in my production workflow for the past 9 months. I'd be more than happy to re-create @bnjt's PR here based off of my fork – @Stranger6667 would you want that?

samgqroberts:
> Hi, @bnjt @platoscave @samgqroberts and others
> 
> I'm working on a model generator from JSON Schema.
> 
> Could you please provide some real world application of such custom fields so I could think how to provide ways to handle such situations in my tool?

In my use case I use a custom string format called "currency", defined as 1 or more digit characters followed by a period followed by 2 digits. This is used as an encoding for currency values to ensure no floating point precision losses, and in lieu of a fixed point data type in JSON. The way that I use the custom keyword validation requires the feature that I added on my fork that allows overriding existing keyworld validation with custom logic, and I do that to override the traditionally numeric keywords `minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`, and `multipleOf` to apply also to strings if the format is "currency".

Stranger6667:
> @platoscave as far as I know this was never merged to the project. I made a fork a while ago that merged @bnjt's work plus an additional piece I needed for my use case I mention in an earlier comment. The last 3 commits of the fork's main branch show that work https://github.com/samgqroberts/jsonschema-rs/commits/master/. For what it's worth I've been using that version of jsonschema-rs in my production workflow for the past 9 months. I'd be more than happy to re-create @bnjt's PR here based off of my fork – @Stranger6667 would you want that?

Yep, it would be amazing to have! :) Thank you

FWIW, I will get back to working on this crate rewrite somewhere mid-April and will incorporate all the new changes in the new version :)

samgqroberts:
@Stranger6667 I've put a copy of this PR + my additional commit up here #473, happy to work with you to get it in

eirnym:
@samgqroberts it's a nice application. If I'd made it myself using only standard schema, I'd use integers (fixed decimal point emulation) for that role or created a string type with a pattern (as floats won't transfer values exactly for all languages) with extra validation in code. I worked in few various projects where the same problem has ben solved in these 2 ways. The only downside of fixed point values is lack of extensibility if you'd want to add more digits than 2 at the start after the decimal dot.

PR #396, opened 2022-11-21
chore: remove deprecated epsilon reference

DanielBauman88:
See https://doc.rust-lang.org/std/f64/constant.EPSILON.html
Updated to https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.EPSILON


Stranger6667:
Thanks!

Issue #397, opened 2022-11-21
multipleOf logic produces incorrect results


DanielBauman88:
The logic for multipleOf compares the remainder of a division with the f64 EPSILON constant to decide if a number is a multiple or not. This approach doesn't make sense. EPSILON is simply the difference between 1.0 and the next larger representable number. With regards to the multipleOf calculation this is just an arbitrary constant and comparing remainders to it gives us no guarantees.

EG:

1e-15 is a multiple of 1e-16. But `(1e-15 / 1e-16) % 1.0 = 1.7763568394002505e-15` which is greater than epsilon so it is incorrectly said to not be a multiple.

1.3e-16 is not a multiple of 1.3. But `(1.3e-16 / 1.3) % 1.0 = 9.999999999999999e-17` which is less than epsilon so it is incorrectly said to be a multiple.

DanielBauman88:
The people over at Monaco have implemented a better way to implement this calculation without introducing incorrect results - https://github.com/microsoft/vscode-json-languageservice/blob/7a478cd29a387cf8813b0a32499e2347102d7a6c/src/parser/jsonParser.ts#L596

Stranger6667:
Will be available in 0.35 with the `arbitrary-precision` feature in the Rust crate + in Python bindings unconditionally, with a caveat - arbitrary precision only works when parsing from JSON strings, not when using language literals:

```python
validator = jsonschema_rs.validator_for('{"multipleOf": 0.1}')
```

Same applies to Rust's `json!()` macro vs `serde_json::from_str()`.

Under the hood, it uses `BigFraction` / `BigInt` when necessary & handles large scientific numbers manually when those types can't handle it.

PR #398, opened 2022-11-21
chore: add test cases to demonstrate multipleOf errors

DanielBauman88:
https://github.com/Stranger6667/jsonschema-rs/issues/397

### DIFF
--- a/jsonschema/src/keywords/multiple_of.rs
+++ b/jsonschema/src/keywords/multiple_of.rs
@@ -169,4 +169,17 @@ mod tests {
     fn schema_path(schema: &Value, instance: &Value, expected: &str) {
         tests_util::assert_schema_path(schema, instance, expected)
     }
+
+    // https://github.com/Stranger6667/jsonschema-rs/issues/397
+    #[test]
+    fn multiple_of_error_cases() {
+        // is not a multiple - but remainder of division is less than epsilon
+        let schema = &json!({"multipleOf": 1.3});
+        let instance = &json!(1.3e-16);
+        tests_util::is_valid(schema, instance);
+        // is a multiple - but remainder of division is greater than epsilon
+        let schema = &json!({"multipleOf": 1e-16});
+        let instance = &json!(1e-15);
+        tests_util::is_not_valid(schema, instance);
+    }
 }


Stranger6667:
Hm, the test is passing

DanielBauman88:
> Hm, the test is passing

It's written to pass so that it can be committed - but those tests logically should fail.

Stranger6667:
Oh, thank you - I completely overlooked it :) I prefer to have this test failing though - so we know that something is not yet properly done. On the other hand, I don't want to overcomplicate things. Maybe, this PR could be a base for porting that implementation from Monako & when it's done we'll have the fixed behavior + regression tests.

Also, maybe this test could also be added to the upstream [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite)?

DanielBauman88:
I could make it fail, but won't that be annoying if every subsequent unrelated PR has failing tests because of this? Would you like me to update this PR with the test that fails?

Could also just cancel this PR, but I think having something making this behaviour clear in the code is nice to have.

DanielBauman88:
> Also, maybe this test could also be added to the upstream

That's a good idea, it will definitely start an interesting conversation. My guess due to the absence of difficult cases in their suite is that the jsonschema people aren't deliberately avoiding dealing with these cases.

I would love for them to take a stand on what they believe is correct behaviour (floating point results vs decimal results).

Stranger6667:
> I could make it fail, but won't that be annoying if every subsequent unrelated PR has failing tests because of this? Would you like me to update this PR with the test that fails?

I was thinking to keep this PR open with failing test cases and then push a fix here as well, and then merge :) What do you think about such a workflow?


I'll open an issue / PR there about this :)

Stranger6667:
Relevant discussion - https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/413

DanielBauman88:
> Relevant discussion - [json-schema-org/JSON-Schema-Test-Suite#413](https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/413)

Yeah, I also started this thread on the slack channel - https://json-schema.slack.com/archives/C5CF75URH/p1669273565065379

For now, the spec doesn't say anything helpful, but for this library we can still choose to do what we think is best for the consumers of this library. I think for most people that is to match decimal results as much as possible.

PR #411, opened 2023-01-29
build(python): Switch to maturin

Stranger6667:


bindings/python/pyproject.toml
```diff
@@ -13,3 +42,11 @@ default_section = "THIRDPARTY"
 include_trailing_comma = true
 known_first_party = "jsonschema_rs"
 known_third_party = []
+
+[tool.maturin]
+python-source = "python"
+strip = true
+
+[build-system]
+requires = ["maturin>=0.14,<0.15"]
```
messense, line 51:
```suggestion
requires = ["maturin>=0.14.11,<0.15"]
```

.github/workflows/build.yml
```diff
@@ -186,10 +186,10 @@ jobs:
           toolchain: stable
           override: true
 
-      - run: python -m pip install tox
+      - run: python -m pip install maturin
         working-directory: ./bindings/python
 
-      - run: tox -e build-sdist
+      - run: maturin build --sdist -o dist
```
messense, line 192: This currently builds both sdist and a wheel, to only build sdist:

```suggestion
      - run: maturin sdist -o dist
```

Stranger6667: Oh, right! Thanks

PR #419, opened 2023-03-15
Add support for `unevaluatedProperties` + a few other changes.

### LINKED ISSUE #288: Implement `unevaluatedProperties`
[Docs](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-11.3) & [example implementation](https://github.com/Julian/jsonschema/blob/main/jsonschema/_validators.py#L430) in Python

tobz:
Closes #288.

This PR adds support for `unevaluatedProperties` when using the `draft201909` or `draft202012` feature flags.

## Overall approach

I took a somewhat similar approach to the Python code linked in the issue for this feature, where the validator for `unevaluatedProperties` uses a subvalidator for each relevant keyword -- `properties`, `patternProperties`, `dependentSchemas`, `$ref`, etc -- that performs the relevant logic for determining if the given schema for the keyword would evaluate the property.

Each subvalidator exposes an interface that looks a lot like the `Validate` trait, but is oriented towards checking solely if a property would be valid against the schema, and has an optional return value to indicate evaluated vs not evaluated. The validator code, overall, is a bit verbose, not unlike `additionalProperties`... but there's a possibility of turning the API into an actual trait which would theoretically allow us to reduce a lot of boilerplate when we check each subvalidator in `UnevaluatedPropertiesValidator::is_valid_property`, and so on.

## Code deduplication and refactoring

Beyond that, I ended up refactoring a little bit of the shared code between `unevaluatedProperties` and `additionalProperties`. I tried, initially, to carry over the logic for switching between a small and big validator map, but it got a little too complex when considering that `UnevaluatedPropertiesValidator` is inherently recursive, and it didn't make a lot of sense propagating a generic validator map type all the way through. I'm happy to try and revisit this if you think it matters enough.

## Fixing `$ref` for draft 2019-09 and newer

In order to fully support `unevaluatedProperties`, I had to do a small bit of refactoring in terms of how `$ref` is handled. This also had the nice side effect of fixing `$ref` usage when using draft 2019-09 and newer, as now it can be evaluated alongside other keywords.

Beyond that, there's just a lot of boilerplate, and I tried to pick a code style/pattern and just stick to that and be consistent about it, but I'm happy to do any of the aforementioned refactoring ideas, or anything else, to bring the code more in line with your expectations.

### DIFF
--- a/jsonschema/src/keywords/unevaluated_properties.rs
+++ b/jsonschema/src/keywords/unevaluated_properties.rs
@@ -1,0 +1,1108 @@
+use std::sync::Arc;
+
+use crate::{
+    compilation::{compile_validators, context::CompilationContext},
+    error::{no_error, ErrorIterator, ValidationError},
+    keywords::CompilationResult,
+    output::BasicOutput,
+    paths::{InstancePath, JSONPointer},
+    primitive_type::PrimitiveType,
+    properties::*,
+    schema_node::SchemaNode,
+    validator::{PartialApplication, Validate},
+};
+use ahash::AHashMap;
+use serde_json::{Map, Value};
+
+/// A validator for unevaluated properties.
+///
+/// In contrast to `additionalProperties`, which can only be used for properties defined in a schema
+/// of type `object`, `unevaluatedProperties` can "see through" advanced validation features like
+/// subschema validation (`allOf`, `oneOf`, `anyOf`), conditional validation (`if`/`then`/`else`),
+/// dependent schemas (`dependentSchemas`), and schema references (`$ref`), which allows applying
+/// `additionalProperties`-like behavior to schemas which use the aforementioned advanced validation
+/// keywords.
+#[derive(Debug)]
+struct UnevaluatedPropertiesValidator {
+    schema_path: JSONPointer,
+    unevaluated: UnevaluatedSubvalidator,
+    properties: Option<PropertySubvalidator>,
+    patterns: Option<PatternSubvalidator>,
+    conditional: Option<Box<ConditionalSubvalidator>>,
+    dependent: Option<DependentSchemaSubvalidator>,
+    reference: Option<ReferenceSubvalidator>,
+    subschemas: Option<Vec<SubschemaSubvalidator>>,
+}
+
+impl UnevaluatedPropertiesValidator {
+    fn compile<'a>(
+        parent: &'a Map<String, Value>,
+        schema: &'a Value,
+        context: &CompilationContext,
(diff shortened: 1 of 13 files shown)


tobz:
Seeing the CI: I'll rebase this locally in a little while to get this using a proper conventional commit message.

jsonschema/src/keywords/unevaluated_properties.rs
```diff
@@ -0,0 +1,1090 @@
...
+/// A subvalidator for unevaluated properties.
+#[derive(Debug)]
+struct UnevaluatedSubvalidator {
+    // The validator associated with `unevaluatedProperties`.
+    //
+    // When `None`, this indicates that `unevaluatedProperties` was `false`, and that no unevaluated
+    // properties are allowed at all. If `Some(None)`, then unevaluated properties _are_ allowed,
+    // and if `Some(Some(...))`, then the contained validator is used to validate any unevaluated
+    // properties.
+    node: Option<Option<SchemaNode>>,
```
Stranger6667, line 561: I am thinking if it would make sense to have a separate enum where each variant will denote each case, instead of using double `Option`. What do you think?

Stranger6667 (commented):
Amazing job, thank you! I have one minor comment :)

jsonschema/src/keywords/unevaluated_properties.rs
tobz: That's fair.

I had sort of the same thought, especially when I was writing the doc comment to explain the behavior of the double `Option<T>`... so I probably should have taken that as a hint in the moment that it should be an enum. :)

tobz: I addressed this in a37e8ca.

Stranger6667 (approved):

Stranger6667:
Great PR and a great write-up! :) Thank you so much for this!

tobz:
@Stranger6667 🙇🏻 Happy to be able to contribute back.

Do you have a typical release cycle for `jsonschema`? Just doing a little planning on my end if I should keep my fork around or not.

tobz:
Great, I'll keep an eye out for when that happens.

Appreciate the timeliness on the review and the release; certainly did not expect my PR to be getting this much traction so quickly. 😂

Stranger6667:
My pleasure :) Happy to see your contribution and that's great that `jsonschema` is getting closer to the full Draft 2019-09 support :)

Stranger6667:
`0.17.0` is released :)

PR #423, opened 2023-03-24
fix: `unevaluatedProperties` doesn't correctly handle subschema validation

### LINKED ISSUE #421: Improper handling of subschema validation in `unevaluatedProperties`.
As part of #419, I chose to make the subschema subvalidator function on a first-match-found basis, such that if you have a `oneOf` with multiple subschemas, the first subschema found that would theoretically evaluate the property _at all_ is the result chosen for the overall validate/apply call on the subvalidator.

However, this is wrong, and leads to unintuitive errors in certain cases. For example, if the schema has `oneOf` where all subschemas are object schemas with a single, identical property, but differing values (`"const": "<value>"`, in my case), then validating that property against any of the subschemas would theoretically result in an application that gave back a result -- because the schema _does_ care about the property -- but may or may not fail depending on the order of the subschemas, and the intended subschema being checked first or not.

In general, we should be obeying the specific behaviors of the given subschema validation mode: continuing with the `oneOf` example, we should be checking all subschemas, and ensuring that only one of them that indicated that it evaluated the property was in fact valid. Similarly, behavior for `allOf` and `anyOf` would need to modeled with their respective constraints.

I'm going to work on a PR to do the aforementioned changes -- treat each subschema validation type (`oneOf`, etc) distinctly -- along with unit tests, now that I can see that I need to be more deliberate about ensuring that validator conforms to the behavior outlined in the specification. This is, I believe, a consequence of the design I implemented -- checking property evaluation directly rather than collecting annotations at the end, like most third-party validators seem to do -- but I do still think it can fixed without much effort and in a clean and internally consistent way.

tobz:
Fixes #421.

## Context

As described in #421, my initial attempt at adding subschema validation handling in `unevaluatedProperties` was a brute-force approach that sufficed to pass the JSON Schema test suite, but didn't correctly evaluate real-world schemas/data.

Instead of only considering properties evaluated if the subschema they were part of was valid, taking into account the behavior of the validator keyword (i.e. `oneOf` can't have multiple matching subschemas), it took an approach of simply finding the first subschema which seemed to favorably evaluate/validate the property.

## Solution

`SubschemaSubvalidator` has been changed in two significant ways:

- it now properly considers the validation type (i.e. `allOf` vs `oneOf`)
- it validates the entire instance before validating the property

In practice, this means that we generate the `SchemaNode` for each subschema, as well as our stripped-down unevaluated properties validator based on the subschema, used for evaluating properties themselves. Using `oneOf` as an example:

- for each subschema in the `oneOf`:
  - validate the property against the property validator (`UnevaluatedPropertiesValidator`) and the property's parent instance against the node validator (`SchemaNode`)
- for each validation result pair:
  - if the instance result was valid, store the property validation result in a local
  - if the local was already populated, return `None` overall: more than one valid subschema in a `oneOf` is not valid
- return the local property validation result

This isn't particularly novel or anything, as it's just following the existing behavior for the `oneOf`/`allOf`/`anyOf` validators, with a little extra logic to thread through the evaluation status of the specific property.

## Testing

I've added three new unit tests -- one for each subschema validation mode -- with fairly simple examples, but they initially did not pass, and now, with this change, they do. 😂 

I've also created a simple CLI program to use `jsonschema` with for taking a schema and input data, and have used it successfully on a large schema document (~800KB, 200+ instances of `unevaluatedProperties: false`, 400-500 schema definitions, etc) where it properly detects... well, if a property is unevaluated or not.

It's still a little weird that the official JSON Schema test suite doesn't have a test for this, but I'm guessing it's based on the fact that their unofficial recommendation for implementing `unevaluatedProperties` is based on checking the output annotations from all of the other keywords, whereas we do it all ourselves in the `unevaluatedProperties` validator. That might be worth thinking about in the future, in terms of refactoring the validator to make it more robust... but for now, I'm just trying to get this working correctly. :)

### DIFF
--- a/jsonschema/src/keywords/unevaluated_properties.rs
+++ b/jsonschema/src/keywords/unevaluated_properties.rs
@@ -1330,10 +1330,25 @@ mod tests {
     use crate::{tests_util, Draft};
     use serde_json::json;
 
+    #[cfg(all(feature = "draft201909", not(feature = "draft202012")))]
+    const fn get_draft_version() -> Draft {
+        Draft::Draft201909
+    }
+
+    #[cfg(all(feature = "draft202012", not(feature = "draft201909")))]
+    const fn get_draft_version() -> Draft {
+        Draft::Draft202012
+    }
+
+    #[cfg(all(feature = "draft201909", feature = "draft202012"))]
+    const fn get_draft_version() -> Draft {
+        Draft::Draft202012
+    }
+
     #[test]
     fn one_of() {
         tests_util::is_valid_with_draft(
-            Draft::Draft201909,
+            get_draft_version(),
             &json!({
                 "oneOf": [
                     { "properties": { "foo": { "const": "bar" } } },
@@ -1348,7 +1363,7 @@ mod tests {
     #[test]
     fn any_of() {
         tests_util::is_valid_with_draft(
-            Draft::Draft201909,
+            get_draft_version(),
             &json!({
                 "anyOf": [
                     { "properties": { "foo": { "minLength": 10 } } },
@@ -1363,7 +1378,7 @@ mod tests {
     #[test]
     fn all_of() {
         tests_util::is_not_valid_with_draft(
-            Draft::Draft201909,
+            get_draft_version(),
             &json!({
                 "allOf": [
                     { "properties": { "foo": { "type": "string" } } },
--- a/jsonschema/src/keywords/unevaluated_properties.rs
+++ b/jsonschema/src/keywords/unevaluated_properties.rs
@@ -1276,7 +1276,7 @@ fn value_has_object_key(value: &Value, key: &str) -> bool {
 fn get_unevaluated_props_schema(parent: &Map<String, Value>) -> &Value {
     parent
         .get("unevaluatedProperties")
-        .unwrap_or(&Value::Bool(false))
+        .unwrap_or(&Value::Bool(true))
 }
 
 pub(crate) fn compile<'a>(
@@ -1379,32 +1379,28 @@ mod tests {
     }
 
     #[test]
-    fn nested_variable_map() {
+    fn all_of_with_additional_props_subschema() {
         tests_util::is_valid_with_draft(
             get_draft_version(),
             &json!({
                 "allOf": [
                     {
-                        "type": "object",
-                        "properties": {
-                            "headers": {
-                                "type": "object",
-                                "additionalProperties": {
-                                    "type": "string"
-                                }
-                            },
-                        },
+                    "type": "object",
+                    "required": [
+                        "foo"
+                    ],
+                    "properties": {
+                        "foo": { "type": "string" }
+                    }
                     },
                     {
                         "type": "object",
-                        "properties": {
-                            "fixed": { "type": "number" }
-                        },
+                        "additionalProperties": { "type": "string" }
                     }
                 ],
                 "unevaluatedProperties": false
             }),
-            &json!({ "fixed": 23, "headers": { "some": "thing" } }),
+            &json!({ "foo": "wee", "another": "thing" }),
         )
     }
 }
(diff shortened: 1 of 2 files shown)


jsonschema/src/keywords/unevaluated_properties.rs
```diff
@@ -1094,6 +1329,11 @@ pub(crate) fn compile<'a>(
     }
 }
 
+fn boxed_errors<'a>(errors: Vec<ValidationError<'a>>) -> ErrorIterator<'a> {
+    let boxed_errors: ErrorIterator<'a> = Box::new(errors.into_iter());
```
Stranger6667, line 1333: Would it work without binding? I.e. immediately return `Box::new(errors.into_iter())`

jsonschema/src/keywords/unevaluated_properties.rs
```diff
@@ -599,15 +779,73 @@ impl SubschemaSubvalidator {
...
+
+            (property_result, instance_result)
+        });
+
+        match self.behavior {
+            // The instance must be valid against _all_ subschemas, and the property must be
+            // evaluated by at least one subschema. We group the errors for the property itself
+            // across all subschemas, though.
+            SubschemaBehavior::All => {
+                let results = mapped.collect::<Vec<_>>();
```
Stranger6667, line 801: Can we reuse `mapped` below without collecting it into a `Vec`?

Stranger6667: P.S. There is a similar case above

jsonschema/src/keywords/unevaluated_properties.rs
tobz: It doesn't, no. The compiler doesn't want to infer that `std::vec::IntoIter<...>` is `dyn Iterator<...>` without it, which is why I made this small helper method in the first place.

Silly compiler. 😠

tobz:
> Maybe we can upstream some of the tests to the official test suite? Also, I guess that @Julian might be interested in reusing some of the logical parts we have in this PR in some other JSON Schema related projects? :) or there could be some useful changes for other implementations too

I actually realized that the commit that the test suite submodule is pointed at is actually fair old, and if I bring it up to the latest commit... a lot of new `unevaluatedProperties` tests are added that don't pass, as well as other, unrelated new tests that also don't pass.

I had the idea that I would get this merged, and then try and open another PR that updates the test suite checkout and fixes whatever needs to be fixed, assuming the fixes are small enough, all in the same PR.

tobz:
As a separate note, using this approach for `unevaluatedProperties` in practice has shown me that it can be quite slow when there is a lot of subschema validation due to the double validation: we run subschema validation ourselves, and then the actual validators for those keywords runs outside of `unevaluatedProperties`.

This is likely why most implementations actually go with the annotation-based approach, examining what properties got evaluated by successful schema nodes, in addition to simplifying the logic for `unevaluatedProperties` itself. I do plan to pursue a refactor to do it with the annotation-based approach, but it would involve -- I believe -- deeper changes to all _other_ property-related keywords to emit the right annotations, so I was going to, again, hold off until I got this implementation working as correctly as possible, and then focus on the "make it fast" part.

Julian:
Hey! Have been a bit behind, sorry. Definitely if there's anything more widely usable I'm interested :)

And yeah certainly so if it turns out there are test gaps even after updating!

tobz:
Really delayed response here (back after an extended work leave) but I'm definitely interested in trying to finally get this merged.

I'll still keep thinking about the aforementioned rework to deal with the performance problems, but this PR is still strictly better from a correctness standpoint, at least. :)

Stranger6667:
@tobz Sorry for the delay from my side too! I agree with you, this PR is a strict improvement over the status quo. I am going to re-review it once more and will work on releasing this :)

tobz:
Makes sense, and glad to see it merged. 👍🏻 

Like I said, I'm still thinking about reworking the support to be in line with other validator implementations, to simplify the code and improve the performance. Hopefully I can get to that in the next few weeks, but, we'll see. :)

PR #436, opened 2023-09-20
fix: Fix multipleOf to work with decimals

torkeln:
A more simplistic approach to validating floats, where we multiply both value and multipleOf with 10 until multipleOf does not have any fractions, and then compare them as integers. This should work since both numbers are fetched from a text string that can't be repeating. Could possibly help solve #397

### DIFF
--- a/jsonschema/src/keywords/multiple_of.rs
+++ b/jsonschema/src/keywords/multiple_of.rs
@@ -9,37 +9,33 @@ use crate::{
 use fraction::{BigFraction, BigUint};
 use serde_json::{Map, Value};
 
-pub(crate) struct MultipleOfFloatValidator {
+pub(crate) struct MultipleOfValidator {
     multiple_of: f64,
     schema_path: JSONPointer,
 }
 
-impl MultipleOfFloatValidator {
+impl MultipleOfValidator {
     #[inline]
     pub(crate) fn compile<'a>(multiple_of: f64, schema_path: JSONPointer) -> CompilationResult<'a> {
-        Ok(Box::new(MultipleOfFloatValidator {
+        Ok(Box::new(MultipleOfValidator {
             multiple_of,
             schema_path,
         }))
     }
 }
 
-impl Validate for MultipleOfFloatValidator {
+impl Validate for MultipleOfValidator {
     fn is_valid(&self, instance: &Value) -> bool {
         if let Value::Number(item) = instance {
-            let item = item.as_f64().expect("Always valid");
-            let remainder = (item / self.multiple_of) % 1.;
-            if remainder.is_nan() {
-                // Involves heap allocations via the underlying `BigUint` type
-                let fraction = BigFraction::from(item) / BigFraction::from(self.multiple_of);
-                if let Some(denom) = fraction.denom() {
-                    denom == &BigUint::from(1_u8)
-                } else {
-                    true
-                }
-            } else {
-                remainder < f64::EPSILON
+            let mut tmp_item = item.as_f64().expect("Always valid");
+            let mut tmp_multiple_of = self.multiple_of;
+
+            while tmp_item.fract() != 0. {
+                tmp_item *= 10.0;
+                tmp_multiple_of *= 10.0;
             }
+
+            tmp_item % tmp_multiple_of == 0.0
         } else {
             true
         }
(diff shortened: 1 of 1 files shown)


OrangeTux:
Pinging @Stranger6667.

Is there any  way we  can bring this issue forward? I'm affected as well by the bug #397. Let me know if I can help.

OrangeTux:
I can confirm the fix works.

OrangeTux:
I've to correct myself. The proposed fix doesn't work for number that contain an exponent.

For example, the current solution correctly identifies that the number `78.0` is a multiple of `0.1`. 
However, it fails the validate that `7.8e1` is a multiple of `0.1`. 

If found this issue by fuzzing with [`cargo fuzz`](https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html). 
Here is a diff to replicate the bug:

```diff
diff --git a/jsonschema/src/keywords/multiple_of.rs b/jsonschema/src/keywords/multiple_of.rs
index 64c1509..3cba1df 100644
--- a/jsonschema/src/keywords/multiple_of.rs
+++ b/jsonschema/src/keywords/multiple_of.rs
@@ -100,6 +100,8 @@ mod tests {
     #[test_case(&json!({"multipleOf": 0.1}), &json!(1.2))]
     #[test_case(&json!({"multipleOf": 0.1}), &json!(1.3))]
     #[test_case(&json!({"multipleOf": 0.02}), &json!(1.02))]
+    #[test_case(&json!({"multipleOf": 0.1}), &json!(78.0))]
+    #[test_case(&json!({"multipleOf": 0.1}), &json!(7.8e1))]
     fn multiple_of_is_valid(schema: &Value, instance: &Value) {
         tests_util::is_valid(schema, instance)
     }
```

OrangeTux:
This code is the core of this PR:

```rust
while tmp_item.fract() != 0. {
    tmp_item *= 10.0;
    tmp_multiple_of *= 10.0;
}
tmp_item % tmp_multiple_of == 0.0
```

Assume `tmp_item` start with `78.0` and `tmp_multiple_of` is `0.1`.
Since `tmp_item.fract() == 0`, the execution jumps to the last line.
Interestingly, `78.0 % 0.1` returns `0.09999884`. Whereas I would expect it returns 0. 
See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=61f390594bccbfee2e853495a30fd52b

OrangeTux:
I found a fix. But I'm not sure yet if the fix works for all inputs.
The fix divides `tmp_item` by `tmp_multiple_of` and validates that the fractional part of the result equals to `0.0`.

```rust
diff --git a/jsonschema/src/keywords/multiple_of.rs b/jsonschema/src/keywords/multiple_of.rs
index 64c1509..eb49b59 100644
--- a/jsonschema/src/keywords/multiple_of.rs
+++ b/jsonschema/src/keywords/multiple_of.rs
@@ -35,7 +35,9 @@ impl Validate for MultipleOfValidator {
                 tmp_multiple_of *= 10.0;
             }

-            tmp_item % tmp_multiple_of == 0.0
+           (tmp_item / tmp_multiple_of).fract() == 0.0
+
+
         } else {
             true
         }
```

OrangeTux:
My solution doesn't work for very large numbers and a value for "mutlipleOf" that's below 1. 
Consider the values 1e308 and 0.1. `1e308 / 0.1 == f32::inf`, and therefore returns `false`.

```rust
diff --git a/jsonschema/src/keywords/multiple_of.rs b/jsonschema/src/keywords/multiple_of.rs
index 64c1509..3d431a8 100644
--- a/jsonschema/src/keywords/multiple_of.rs
+++ b/jsonschema/src/keywords/multiple_of.rs
@@ -100,6 +100,8 @@ mod tests {
     #[test_case(&json!({"multipleOf": 0.1}), &json!(1.2))]
     #[test_case(&json!({"multipleOf": 0.1}), &json!(1.3))]
     #[test_case(&json!({"multipleOf": 0.02}), &json!(1.02))]
+    #[test_case(&json!({"multipleOf": 0.5}), &json!(1e308))]
     fn multiple_of_is_valid(schema: &Value, instance: &Value) {
         tests_util::is_valid(schema, instance)
     }

OrangeTux:
Here's a relevant discussion explaining why `78.0 % 0.1` is not `0.0`: https://users.rust-lang.org/t/how-does-the-operator-work-with-floating-point-numbers/82030

Stranger6667:
Closing in favor of #750 - test cases from this PR pass now.

P.S. Sorry for the absence of communication from my side here

PR #437, opened 2023-09-20
Build aarch64 Python wheels

### LINKED ISSUE #431: python linux aarch64/arm64
it would be nice to publish wheels for this architecture.

kgutwin:
This updates the `python-release.yml` Actions workflow to build and publish wheels for ARM (aarch64 / amd64).

Note that the new "Install built wheel on ARM architecture" step could be omitted; it is only there to demonstrate that the wheel will install in the appropriate architecture.

Resolves #431.

Thanks!


Negaihoshi:
This solution can also help me, and I hope we can merge it.

PR #447, opened 2024-01-15
chore: Fix typos

jayvdb:


jsonschema/src/error.rs
```diff
@@ -927,14 +927,14 @@ impl fmt::Display for ValidationError<'_> {
             ),
             ValidationErrorKind::MaxProperties { limit } => write!(
                 f,
-                "{} has more than {} propert{}",
+                "{} has more than {} property{}",
```
Stranger6667, line 930: This one is conditional, see the last argument to the macro

jayvdb: reverted that

Stranger6667 (approved):
Thanks!

PR #457, opened 2024-03-01
Expose drafts 2019-09 and 2020-12 to Python

DerRidda:
Very small PR that simply exposes the already supported but not yet exposed drafts 2019-09 and 2020-12 to Python.
Not really tested because the project right now doesn't seem to build cleanly from a fresh checkout.


Stranger6667:
FYI @DerRidda 

Released in `0.17.2`

PR #466, opened 2024-03-21
feat(python): support validation of dict subclasses

### LINKED ISSUE #427: [python] ValueError: Unsupported type: 'collections.OrderedDict'
Even though `OrderedDict` is a subclass of `dict`, passing it to `jsonschema_rs` functions raises an exception:
```python
>>> import collections
>>> import jsonschema_rs
>>> jsonschema_rs.is_valid({"type": "object"}, collections.OrderedDict({}))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Unsupported type: 'collections.OrderedDict'
```
How hard would it be to support subclasses of `dict`?

flbraun:
Disclaimer: This is a low-confidence PR as I have zero experience with Rust and the pyo3 library. Feel free to be hard on me :)

Currently, attempting to validate an instance of a dict subclass will raise a `ValueError`. This should not be happening, since the instance is still dict. Achieve compatibility by checking the subclasses' inheritance tree, and treat the instance like a dict if that check passes.

Resolves #427

### DIFF
--- a/bindings/python/CHANGELOG.md
+++ b/bindings/python/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Support subclasses of Python `dict`s [#427](https://github.com/Stranger6667/jsonschema-rs/issues/427)
+
 ## [0.17.2] - 2024-03-03
 
 ### Added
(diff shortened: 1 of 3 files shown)


Stranger6667:
Thank you! It looks good to me, will check it in more details during the weekend

Stranger6667 (approved):

Stranger6667:
Thank you so much! I'll issue a new release shortly

PR #473, opened 2024-04-10
feat: Custom keyword validation

### LINKED ISSUE #379: Custom Validators

samgqroberts:
This was originally put up by @bnjt here https://github.com/Stranger6667/jsonschema-rs/pull/394. That repository was deleted which deleted the PR, so this PR re-creates that PR.

Additionally, this PR adds an improvement to the original PR whereby existing keywords can also have their behavior be overridden by custom logic.

Resolves #379
Resolves #354


jsonschema/src/keywords/custom_keyword.rs
```diff
@@ -0,0 +1,71 @@
+use crate::compilation::context::CompilationContext;
+use crate::compilation::options::CustomKeywordDefinition;
+use crate::keywords::CompilationResult;
+use crate::paths::{InstancePath, JSONPointer, PathChunk};
+use crate::validator::Validate;
+use crate::ErrorIterator;
+use serde_json::{json, Value};
+use std::fmt::{Display, Formatter};
+use std::sync::Arc;
```
Stranger6667, line 9: I got to set `rustfmt.toml` to make sure everything is consistently formatted, but generally the imports are grouped in this crate.

jsonschema/src/keywords/custom_keyword.rs
```diff
@@ -0,0 +1,71 @@
...
+
+// An Arc<Value> is used so the borrow checker doesn't need explicit lifetime parameters.
+// This would pollute dependents with lifetime parameters.
+pub(crate) type CustomValidateFn =
```
Stranger6667, line 13: I see that the `CustomIsValidFn` function uses regular references, do we need `Arc` here? I.e. how would the signature look like otherwise as you mention extra lifetimes

jsonschema/src/compilation/mod.rs
```diff
@@ -302,4 +320,253 @@ mod tests {
...
+                return false; // Invalid schema
+            }
+            for (key, _value) in instance.as_object().unwrap() {
+                if !key.is_ascii() {
+                    return false;
+                }
+            }
+            true
+        }
+        let definition = CustomKeywordDefinition::Validator {
```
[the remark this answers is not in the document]
samgqroberts: That sounds like a good idea, but a few clarifying questions: do you mean creating a trait like `CustomValidator` with the `validate` and `is_valid` functions as methods on it, and then the user would define a struct that implements `CustomValidator` and provide it to CompilationOptions (in like a Box I think)? If so perhaps it should be combined somehow with the `Validator` trait?

Also while instinctually this feels like the right move, I'm struggling a bit to come up with a concrete use case example, like for a test case. There's this https://github.com/Stranger6667/jsonschema-rs/pull/394#issuecomment-2047665476, but I wonder if that doesn't need state _internal_ to the validator, the `validate` / `is_valid` functions could potentially just reference the lookup table. Something like that's probably worth a test case anyway also.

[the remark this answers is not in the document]
Stranger6667: > do you mean creating a trait like CustomValidator with the validate and is_valid functions as methods on it, and then the user would define a struct that implements CustomValidator and provide it to CompilationOptions (in like a Box I think)? If so perhaps it should be combined somehow with the Validator trait?

Yep, or, maybe exposing the `Validator` trait?

> Also while instinctually this feels like the right move, I'm struggling a bit to come up with a concrete use case example, like for a test case

An example that I have in mind is a counter. Even though it might sound trivial, it unblocks interesting use cases like collecting usage statistics from different keywords (i.e. if the user "wraps" the existing keyword implementation with such a counter), which in turn is useful for building coverage maps for test cases send to API endpoints (documented with Open API), which in turn is useful for coverage-guided fuzzing of such endpoints.

jsonschema/src/compilation/mod.rs
```diff
@@ -569,4 +582,58 @@ mod tests {
...
+                };
+                self.count += amount as u32;
+                true
+            }
+        }
+
+        // define compilation options that include the custom format and the overridden keyword
+        let validator = Arc::new(Mutex::new(Box::new(CountingValidator { count: 0 })));
+        let mut options = JSONSchema::options();
+        let options = options.with_custom_keyword("countme", validator.clone());
```
samgqroberts, line 623: @Stranger6667 unfortunately I think I've hit the end of my expertise with trait objects, Arc, Mutex, Box, etc. Everything else compiles, however I can't figure out how to maintain the type of `validator` here so that I can access the `count` member later on in assertions.

```
error[E0308]: mismatched types
   --> src/compilation/mod.rs:623:62
    |
623 |         let options = options.with_custom_keyword("countme", validator.clone());
    |                               -------------------            ^^^^^^^^^^^^^^^^^ expected `Arc<Mutex<Box<...>>>`, found `Arc<Mutex<Box<CountingValidator>>>`
    |                               |
    |                               arguments to this method are incorrect
    |
    = note: expected struct `Arc<std::sync::Mutex<Box<(dyn for<'instance, 'schema> CustomKeywordValidator<'instance, 'schema> + 'static)>>>`
               found struct `Arc<std::sync::Mutex<Box<CountingValidator>>>`
    = help: `CountingValidator` implements `CustomKeywordValidator` so you could box the found value and coerce it to the trait object `Box<dyn CustomKeywordValidator>`, you will have to change the expected type as well```
    
Also, I'm not thrilled with how the `with_custom_keyword` interface requires an Arc of a Mutex of a Box of the thing... for non inner state use cases that's unnecessary (see the other custom keyword tests above), but I'm not sure how to support both. Any ideas for either problem?

Comment (Stranger6667) [reply jsonschema/src/compilation/mod.rs:623]: Looking at it right now. I'll push a few commits to your branch which will this work and then we can continue discussing the interface :)

File: jsonschema/src/compilation/mod.rs [hunk 5]
```diff
@@ -588,51 +583,63 @@ mod tests {
...
-                self.count += amount as u32;
+                self.increment(amount);
                 true
             }
         }
 
         // define compilation options that include the custom format and the overridden keyword
-        let validator = Arc::new(Mutex::new(Box::new(CountingValidator { count: 0 })));
+        let count = Arc::new(Mutex::new(0));
+        let boxed: Box<dyn CustomKeywordValidator> = Box::new(CountingValidator {
```
samgqroberts, line 625: It's interesting that Rust can't preserve that this is a `CountingValidator` while passing it to a function that expects a `dyn CustomKeywordValidator`. That's what I was getting hung up on. The inner state being behind another Arc<Mutex<T>> seems like a good alternative.

jsonschema/src/compilation/mod.rs
Stranger6667: Here are a few thoughts:

Given the inner state, I think that `custom_keywords` should store initialization functions (boxed this time, unlike `formats`), so such validators won't be shared across multiple locations in the schema. This will remove the need for `Arc<Mutex<T>>`, so no unconditional locking when any custom validator is used - stateless validators will be more performant. Alternatively, we can store something that implements `Validator + Default`, which is likely a better option.

jsonschema/src/compilation/mod.rs
Stranger6667: Technically, Rust preserves this information, but behind an indirection. Dynamic dispatch "knows" at runtime what kind of object is there, so it can do a vtable lookup and find the relevant implementation of methods from `CustomKeywordValidator`. It is also possible to downcast `dyn CustomKeywordValidator` to `CountingValidator` via `downcast_ref`, though it is not that convenient to use.

Without this kind of indirection, it would not be clear how much space to allocate on the stack for that type + as we store it in a hashmap, we can only use boxed trait objects.

jsonschema/src/keywords/custom_keyword.rs
Stranger6667: Btw, I merged #474 so, `rustfmt` now checks for such cases

Stranger6667:
Let me know if there is anything I can do to help with moving this forward :)

samgqroberts:
@Stranger6667 sounds good! day job has taken over this week but I'll dig back into this PR this weekend

jsonschema/src/compilation/mod.rs
samgqroberts: Hey, so, I think again my inexperience is showing. My attempts with Validator + Default fail because that cannot be a trait object, and if we try to store a fn in CompilationOptions we can't capture variables when in the closure and that wouldn't be able to clone the `count` in the inner state test, and if we try to store a Fn then that's either not thread safe or not cloneable. Any ideas?

Stranger6667: Oh, I haven't thought about the trait object thing. Let me play with the code a bit

P.S. Offtopic - I've put together some ideas about this library's `1.0` API and would appreciate any feedback (see https://github.com/Stranger6667/jsonschema-rs/discussions/475)

Stranger6667: Done in a240bd5aae1903b00328a444046dc7b008793928! However, there are a few issues:

- Initialization functions require `&'static` so, it is problematic to pass a value there (like a counter) and then read from it without leaking.
- The API does not spark joy :( However, given that there could be multiple occurrences of the same keyword, every one of them should have its separate state, so it is probably the only way to provide this kind of capability. I.e. - each validator is unique, so we need to give a way to initialize each of them. With stateless validators, things could be simpler.

Not 100% sure if it is worth the trouble and if the use case I mentioned above could be implemented differently

Stranger6667:
I have an idea how what the API could look like:

```rust
#[derive(Debug)]
struct AsciiKeyword {
    size: usize
}

impl jsonschema::CustomKeyword for AsciiKeyword {
    fn is_valid(&self, instance: &Value) -> bool {
        todo!()
    }
}

fn ascii_keyword(schema: &Value) -> Arc<dyn jsonschema::CustomKeyword> {
    Arc::new(AsciiKeyword { size: 42 })
}

let validator = JSONSchema::options()
    .with_keyword(
        "ascii",
        |schema| -> Arc<dyn jsonschema::CustomKeyword> {
            Arc::new(AsciiKeyword { size: 42 })
        }
    )
    .keyword("also-ascii", ascii_keyword)
    .compile(&schema)
    .unwrap();
```

I've implemented it for the new `jsonschema` version - at least it compiles and the code above works. Here is the implementation:

- [Traits](https://github.com/Stranger6667/jsonschema/blob/main/src/vocabulary/mod.rs#L14-L46) - `CustomKeyword` that provides the relevant methods (`is_valid`, etc), `CustomKeywordFactory` is what is actually stored inside `options`, and is implemented for functions that accept 1 argument of the `Value` type. Those should be extended to match all the needed arguments from the current implementation
- [Function to add new keywords](https://github.com/Stranger6667/jsonschema/blob/main/src/validation/builder.rs#L53-L60)

With some adjustments (mainly for other arguments like path, etc) it should work

samgqroberts:
@Stranger6667 Thanks for running with this. I'll be ready to take another look after your updates.

Stranger6667:
I'll add some more changes soon! Here are a few observations:

- Some logic should be moved to the "build" phase, so as soon as we know that the schema is invalid it should be reported. Hence the keyword factory should return `Result`
- Should custom validators return a single error? It may simplify the logic a bit

Otherwise, I think that the path forward is more or less clear

Stranger6667:
Update:

I think that the API is more or less done (and except a few more args) and I am going to adjust the tests today/tomorrow. Not sure if closures would work as keyword factories (without a lifetime on trait I don’t know how to make them work) without breaking changes to the “options” type. But regular functions are OK.

so, I’ll do some extra cleanup and will update the docs. Hopefully it should be enough to cover the usecase you have and reduce the memory/performance overhead in comparison with the original approach

Stranger6667 (approved):

samgqroberts:
👏 Thank you @Stranger6667 – the final draft looks great for my use case

Stranger6667:
Thank you @samgqroberts ! Let me know if you need anything else from my side :) I’d be happy to chat

PR #487, opened 2024-05-21
feat: configure patterns regex engine

shumkov:
Fancy Regex supports backtracking which is required for some cases but as a downside is vulnerable to ReDoS attacks. This becomes a decisive factor when an application operates with user-defined schemas. Regex, in turn, doesn't support look-around and backreferences but guarantees linear time matching that mitigates the attack.

This PR enables the configuration of the regex engine for pattern-based keywords: `Regex` or `FancyRegex` (by default).

```rust
use jsonschema::{CompilationOptions, RegexEngine, RegexOptions};
let mut options = CompilationOptions::default();
// Set Regex as a default engine for pattern keyword
options.with_patterns_regex_engine(RegexEngine::Regex(RegexOptions {
  size_limit: Some(5 * (1 << 20)),
  ..Default::default()
}));
```

The formats still use Fancy Regex. I didn't find a simple way to keep patterns static and configurable at the same time. Probably, the right approach is to add an option to use fast formats such as [ajv-formats](https://github.com/ajv-validator/ajv-formats) but this is out of the scope of this PR.

### DIFF
--- a/jsonschema/src/keywords/format.rs
+++ b/jsonschema/src/keywords/format.rs
@@ -1,7 +1,6 @@
 //! Validator for `format` keyword.
 use std::{net::IpAddr, str::FromStr, sync::Arc};
 
-use fancy_regex::Regex;
 use once_cell::sync::Lazy;
 use serde_json::{Map, Value};
 use url::Url;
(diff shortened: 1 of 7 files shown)


shumkov:
Hello @Stranger6667! Will it be interesting for you guys or it's just my specific use case?

Stranger6667:
This looks cool! Sorry for the delay, I’ll check it in detail in the next couple of days and will let you know

jsonschema/src/keywords/format.rs
```diff
@@ -281,12 +281,28 @@ impl Validate for JSONPointerValidator {
         }
     }
 }
-format_validator!(RegexValidator, "regex");
+struct RegexValidator {
+    schema_path: JSONPointer,
+    config: Arc<CompilationOptions>,
```
Stranger6667, line 286: Maybe keeping the regex engine options will be enough, instead of storing the whole set of options? Thinking more about just storing the necessary minimum, rather than performance. However, I think as cloning `RegexEngine` is cheap, it could be also better performance wise during validation (though for a cost of a pointer indirection)

Stranger6667 (changes requested):
I like the changes! One minor thing re options

Also, please, add a changelog entry :)

Also, maybe instead of 

```rust
options.with_patterns_regex_engine(RegexEngine::Regex(RegexOptions {
  size_limit: Some(5 * (1 << 20)),
  ..Default::default()
}));
```

It could be

```rust
options.with_patterns_regex_engine(RegexOptions {
  size_limit: Some(5 * (1 << 20)),
  ..Default::default()
});
```

?

It could work if `with_patterns_regex_engine` will accept a generic that can do `Into<RegexEngine>`, so implementing `Into` will make the API a bit more ergonomic.

P.S. `cargo fmt` & `clippy` are complaining, but it should be straightforward to fix

Stranger6667:
This feature will be available in `0.30.0`, sorry for the delay

Issue #496, opened 2024-09-09
Incorrect/inconsistent validation result when combining `unevaluatedProperties`/`allOf`/`oneOf`


GREsau:
Thanks for the crate, I've started using this to reimplement the tests in schemars to validate its outputs, and doing so has already revealed a few schemars bugs!

I also stumbled upon this odd case though - full repro:

```rust
use jsonschema::{Draft, JSONSchema};
use serde_json::json;

fn main() {
    let schema = JSONSchema::options()
        .with_draft(Draft::Draft202012)
        .compile(&json!({
            "$schema": "https://json-schema.org/draft/2020-12/schema",
            "allOf": [{}],
            "oneOf": [
                {
                    "properties": {
                        "blah": true
                    }
                }
            ],
            "unevaluatedProperties": false
        }))
        .unwrap();

    let value = json!({
        "blah": 1
    });

    println!(
        "Is valid: {}, validate is OK: {}",
        schema.is_valid(&value),
        schema.validate(&value).is_ok()
    );
}
```

I believe that schema is functionally equivalent to
```json
{
  "properties": {
    "blah": true
  },
  "additionalProperties": false
}
```

So I'd expect the output to be `Is valid: true, validate is OK: true`, but it's actually `Is valid: false, validate is OK: true`. i.e. `validate` correctly passes validation, but `is_valid` fails it for some reason

Stranger6667:
> Thanks for the crate, I've started using this to reimplement the tests in schemars to validate its outputs, and doing so has already revealed a few schemars bugs!

Delighted to hear it!

Indeed, it is a bug! Thanks for reporting :)

PR #497, opened 2024-09-09
Ignore trailing '#' in meta-schema URLs

GREsau:
A single trailing # should be present for draft-07 and earlier, but **should not** be present for 2019-09 or 2020-12 (e.g. see the `$schema` value in [draft2020-12/schema.json](https://github.com/Stranger6667/jsonschema-rs/blob/master/jsonschema/meta_schemas/draft2020-12/schema.json)). But for simplicity and robustness, we can just ignore it in all cases.

With this change, the output of this example is now `true`:
```rust
let schema = JSONSchema::compile(&json!({
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "unevaluatedProperties": false
}))
.unwrap();

let value = json!({
    "blah": 1
});

println!("{}", schema.is_valid(&value));
```

### DIFF
--- a/jsonschema/src/schemas.rs
+++ b/jsonschema/src/schemas.rs
@@ -184,14 +184,14 @@ impl Draft {
 /// Get the `Draft` from a JSON Schema URL.
 #[inline]
 pub(crate) fn draft_from_url(url: &str) -> Option<Draft> {
-    match url {
+    match url.trim_end_matches('#') {
         #[cfg(feature = "draft202012")]
-        "https://json-schema.org/draft/2020-12/schema#" => Some(Draft::Draft202012),
+        "https://json-schema.org/draft/2020-12/schema" => Some(Draft::Draft202012),
         #[cfg(feature = "draft201909")]
-        "https://json-schema.org/draft/2019-09/schema#" => Some(Draft::Draft201909),
-        "http://json-schema.org/draft-07/schema#" => Some(Draft::Draft7),
-        "http://json-schema.org/draft-06/schema#" => Some(Draft::Draft6),
-        "http://json-schema.org/draft-04/schema#" => Some(Draft::Draft4),
+        "https://json-schema.org/draft/2019-09/schema" => Some(Draft::Draft201909),
+        "http://json-schema.org/draft-07/schema" => Some(Draft::Draft7),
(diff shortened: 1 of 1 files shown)


Stranger6667:
Thanks!

Should the example be evaluated as `true`? Given the [example](https://www.learnjsonschema.com/2020-12/unevaluated/unevaluatedproperties/#examples) from `learnjsonschema.com`:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "unevaluatedProperties": false
}
```

> All object instances fail against the false schema

Am I correct that this change exposes #496 more clearly?

GREsau:
> Should the example be evaluated as true?

Whoops, sorry that was a typo. I meant it currently outputs true, but with this PR, it now outputs false!

GREsau:
> Am I correct that this change exposes https://github.com/Stranger6667/jsonschema-rs/issues/496 more clearly?

Kinda, in that you no longer need to specify `.with_draft(Draft::Draft202012)`, as it will now infer that it's a 2020-12 schema from `$schema`

Stranger6667:
> Whoops, sorry that was a typo. I meant it currently outputs true, but with this PR, it now outputs false!

Aha, got it! So, the other part of my comment is irrelevant then :) I haven't checked it myself yet, but it would not fix #496, right? as it seems to me that #496 is caused by some issue in [UnevaluatedPropertiesValidator::is_valid](https://github.com/Stranger6667/jsonschema-rs/blob/f001fe6534ffaf704cada8be20daefda3cc7c6cb/jsonschema/src/keywords/unevaluated_properties.rs#L343)

Stranger6667 (approved):

GREsau:
> I haven't checked it myself yet, but it would not fix #496, right? as it seems to me that #496 is caused by some issue in [UnevaluatedPropertiesValidator::is_valid](https://github.com/Stranger6667/jsonschema-rs/blob/f001fe6534ffaf704cada8be20daefda3cc7c6cb/jsonschema/src/keywords/unevaluated_properties.rs#L343)

Correct, they're entirely separate issues

Stranger6667:
Good. I am going through the failures in [bowtie.report](https://bowtie.report/#/implementations/rust-jsonschema) and will push some fixes this week. For today, I plan to fix the issue of `$ref` ignoring `$schema` inside the resolved schema.

Issue #451, opened 2024-02-05
Implement `core::error::Error` for `ErrorIterator`


tgross35:
To do this `ErrorIterator` needs a `Display` impl, which could require making it a wrapper over `Box<dyn Iterator<Item = ValidationError<'a>> + Sync + Send + 'a>`, rather than just a type alias (a mild breaking change).

This will allow throwing this error to `dyn Error` or `anyhow` without any manual mapping. This impl should probably just print a list of all errors, like the examples show. Roughly:

```rust
type ErrorIterTy<'a> = Box<dyn Iterator<Item = ValidationError<'a>> + Sync + Send + 'a>;
pub struct ErrorIterator<'a>(ErrorIterTy<'a>);

impl<'a> IntoIterator for ErrorIterator<'a> {
    type Item = ValidationError<'a>;
    type IntoIter: ErrorIterTy<'a>;

    fn into_iter(self) -> Self::IntoIter {
        self.0
    }
}

impl fmt::Display for ErrorIterator {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
        writeln!(f, "Validation errors:");
        for (idx, e) in self.enumerate() {
            writeln!(f, "{idx:02}: {e}")?;
        }
        Ok(())
    }
}

impl core::error::Error for ErrorIterator {}
```

Stranger6667:
Hey, sorry for the delay!

First of all, thank you for opening this issue, it has been unaddressed way too long. 

My inclination is to adjust the API so this type is only returned by a new `iter_errors` function, and `validate` will return just the first error. I believe it will make the API more consistent with other libraries in the JSON Schema ecosystem (e.g. Python's `jsonschema` has it + Python bindings to this crate already have it). Even though it is a significant breaking change, I think it should be addressed before `1.0`.

Other than that, I completely agree with the proposed changes and will address it in the next couple of releases.

PR #613, opened 2024-10-24
refactor: Rework iter_errors & validate

Stranger6667:
This PR introduces a few important breaking changes to the public API:

- Change the return type of the `Validator::validate` to `Result<(), ValidationError<'i>>`
- Add `Validator::iter_errors` that returns `ErrorIterator<'i>`

Which is essentially a split of the existing `Validator::validate` methods into two methods, one of which returns the first error, and the second one allows iterating over all errors.

Besides the performance improvements for cases when the user needs just the first error (Python bindings benefit a lot from it), it is also a step toward a clearer API. 

I see the new API as less surprising because:
- with the new `validate` methods, the user won't have to check the first value of the iterator to get the error - it is there straight away. 
- If the user needs all errors, then iteration does not involve unwrapping `Result`, which I think is desirable

Another pending change is to implement #451, but I am not 100% about the error formatting inside the iterator `Display`

TODO:
- [x] Migration guide
- [x] Documentation
- [x] Top-level `jsonschema::validate` function.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
       - run: cargo fmt --all -- --check
 
   lints-rust:
-    name: Rust lints on WASM
+    name: Rust lints
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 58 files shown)


denzalman:
can you pls update documentation? It's very not straight forward what to do with this bracing change without reading gh issues
image

Stranger6667:
@denzalman 

Thank you for pinging! Sorry, I've missed it - will update soon

PR #624, opened 2024-10-26
perf: short-circuit SchemaNode.is_valid

jqnatividad:
there's an optimization already for just one validator, perhaps, we should short-circuit as well as soon as one validator is not valid

### DIFF
--- a/crates/jsonschema/src/node.rs
+++ b/crates/jsonschema/src/node.rs
@@ -334,7 +334,12 @@ impl Validate for SchemaNode {
                 kvs.validators[0].1.is_valid(instance)
             }
             NodeValidators::Keyword(kvs) => {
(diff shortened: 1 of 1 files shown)


Stranger6667:
Thanks!

The `all` method [short-circuits](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all):

> all() is short-circuiting; in other words, it will stop processing as soon as it finds a false, given that no matter what else happens, the result will also be false.

The manual version generates somewhat [less assembly](https://godbolt.org/z/fWK4zEbjc) (conceptually similar code) which at first glance has roughly the same instructions. I didn't check the actual details, but I think the gain is marginal as this difference is dominated by other factors.

I did something similar in `uniqueItems`, so I think it is reasonable to accept it here too

PR #652, opened 2024-12-22
feat(python): Custom retrievers for external references

### LINKED ISSUE #372: How to use custom resolvers in python?
I have JSON schema splitted to more than one file (with refs) in local filesystem.

I use in python: `jsonschema_rs.JSONSchema.from_str` and then `validate`.

Result:

>failed to resolve json-schema:///myschema.schema.json: cannot resolve relative external schema without root schema ID

It isn't surprising because `jsonschema_rs` isn't aware where to look for this external reference.

Is it possible to use [custom resolver](https://github.com/Stranger6667/jsonschema-rs/pull/341) when using `jsonschema_rs` from python?

Stranger6667:
Resolves #372

### DIFF
--- a/crates/jsonschema-py/CHANGELOG.md
+++ b/crates/jsonschema-py/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Custom retrievers for external references. [#372](https://github.com/Stranger6667/jsonschema/issues/372)
+
 ### Changed
 
 - Improved error message for unknown formats.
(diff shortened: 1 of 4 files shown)


Stranger6667:
cc @jpmckinney I think it should be enough for your use case, right? If not, please, let me know and I'll adjust the behavior in a follow-up PR.

jpmckinney:
That was fast! Yes, I can use that both to control which URI schemes are supported, and I can use it like a registry (mapping individual URLs to specific values).

PR #683, opened 2025-01-29
Ensure license file is published with each crate

a-wai:
As the license file is present at the root of the repository, it isn't part of any crate on crates.io, as those lie in subfolders. Ensure this file is published by telling cargo where to find it.


Stranger6667 (approved):

Stranger6667:
Thank you!

a-wai:
Oh, I wasn't aware of that, my bad! FWIW I was inspired by [this PR](https://github.com/probe-rs/rtt-target/pull/29), but I'll keep this in mind for the future.

I guess a better fix would be for crates.io or cargo to add a LICENSE file when there is none but `license` is set and valid, I'll check if/what I can do there.

PR #703, opened 2025-02-07
feat: Async retriever

### LINKED ISSUE #385: Async SchemaResolver
I'm using an async S3 client to load schemas from a bucket.  I've hit a wall in that SchemaResolver is not async, but I need to call an async method within it.  Is AsyncSchemaResolver on the roadmap or is there a known work around?

```rust
///
/// Download an object from the S3 bucket and turn the result as Bytes
/// 
pub async fn download_object(client: &Client, bucket_name: &str, key: &str) -> Result<Bytes, S3Error> {
    let resp = client
        .get_object()
        .bucket(bucket_name)
        .key(key)
        .send()
        .await?;
    let data = match resp.body.collect().await {
        Ok(data) => data,
        Err(e) => return Err(S3Error::GetObjectError{ bucket_name: bucket_name.to_string(), key: key.to_string(), msg: e.to_string() }),
    };
    let data = data.into_bytes();

    Ok(data)
}

///
/// Download and object from the S3 bucket and return the results as a json Value
/// 
pub async fn download_json(client: &Client, bucket_name: &str, path: &str) -> Result<serde_json::Value, JsonError> {
    let data = download_object(client, bucket_name, path).await?;
    let json: serde_json::Value = serde_json::from_slice(&data)?;
    Ok(json)
}

struct S3SchemaResolver {
    client: Client,         // S3 client
    bucket_name: String     // name of bucket that holds schemas
}

impl SchemaResolver for S3SchemaResolver {
    fn resolve(&self, root_schema: &Value, url: &Url, original_reference: &str) -> Result<Arc<Value>, SchemaResolverError> {
        //
        // add the base path to the reference
        //
        match download_json(&self.client, &self.bucket_name, original_reference).await {
            Ok(schema_value) => Ok(Arc::new(schema_value)),
            Err(json_error) => {
                let msg = format!("{}", json_error);
                log::error!("{}", msg);
                Err(anyhow::anyhow!(msg))
            }
        }
    }
}
```

Stranger6667:
Resolves #385 

```rust
let validator = jsonschema::async_validator_for(&schema).await?;
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           cache-all-crates: "true"
           key: ${{ matrix.os }}
 
-      - run: cargo test --no-fail-fast
+      - run: cargo test --no-fail-fast --all-features
 
   test-wasm:
     name: Test on WASM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,39 @@
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- All builder methods on `ValidationOptions` now take ownership of `self` instead of `&mut self`.
+  This change enables better support for non-blocking retrieval of external resources during the process of building a validator.
+  Update your code to chain the builder methods instead of reusing the options instance:
+
+  ```rust
+  // Before (0.28.x)
+  let mut options = jsonschema::options();
+  options.with_draft(Draft::Draft202012);
+  options.with_format("custom", my_format);
+  let validator = options.build(&schema)?;
+
+  // After (0.29.0)
+  let validator = jsonschema::options()
+      .with_draft(Draft::Draft202012)
+      .with_format("custom", my_format)
+      .build(&schema)?;
+
+- The `Retrieve` trait's `retrieve` method now accepts URI references as `&Uri<String>` instead of `&Uri<&str>`.
+  This aligns with the async version and simplifies internal URI handling. The behavior and available methods remain the same, this is purely a type-level change.
+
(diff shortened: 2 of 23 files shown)


Stranger6667:
Tagging @uncenter @gadomski because of your recent PRs (in `taplo` and `stac-rs` respectively) which may benefit from this feature in `jsonschema`. 

This is a very early iteration for async retrieval of external schemas, but it works (the original message with an example above)! If you have any feedback on the API or think that I've missed something here, please, let me know :) 

I plan to work more on this feature so it covers your use cases and has the least amount of rough edges in terms of usability and ergonomics.

Stranger6667:
With the latest update, the following would work for the simplest case:

```rust
let validator = jsonschema::async_validator_for(&schema).await?;
```

And custom non-blocking retrievers are also possible:

```rust
use serde_json::{Value, json};
use jsonschema::{Draft, AsyncRetrieve, Uri};

// Custom async retriever
struct MyRetriever;

#[async_trait::async_trait]
impl AsyncRetrieve for MyRetriever {
    async fn retrieve(&self, uri: &Uri<String>) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
        // Custom retrieval logic
        Ok(json!({}))
    }
}

let schema = json!({
    "$ref": "https://example.com/user.json"
});
let validator = jsonschema::async_options()
    .with_draft(Draft::Draft202012)
    .with_async_retriever(MyRetriever)
    .build(&schema)
    .await?;
```

I hope it should be enough for the majority of use cases

PR #716, opened 2025-03-22
fix(python): added missing Registry to type declarations

johnnoone:
The new class `Registry` were missing from  type declarations.
This MR adds this into the `.pyi` file


Stranger6667:
Thanks!

Issue #675, opened 2025-01-22
Memory not being freed after validation


PrbNoblem:
After performing seemingly any type of json schema validation, it seems that memory is not being freed. For example, running this version of the example code with the [cap](https://crates.io/crates/cap) crate to print memory usage:

```
use serde_json::json;
use std::alloc;
use cap::Cap;

#[global_allocator]
static ALLOCATOR: Cap<alloc::System> = Cap::new(alloc::System, usize::max_value());

fn main() -> Result<(), Box<dyn std::error::Error>> {
    println!("Allocated before validation: {}KB", ALLOCATOR.allocated()/1024);
    {
        let schema = json!({"maxLength":5});
        let instance = json!("foo");
        let validator = jsonschema::validator_for(&schema)?;
        assert!(validator.validate(&instance).is_ok());
    }
    println!("Allocated after validation: {}KB", ALLOCATOR.allocated()/1024);
    Ok(())
}
```

results in the following being printed
```
Allocated before validation: 0KB               
Allocated after validation: 4184KB
```
Using the jsonschema::is_valid approach instead results in the same behavior:
```
println!("Allocated before validation: {}KB", ALLOCATOR.allocated()/1024);
    {
        let schema = json!({"maxLength":5});
        let instance = json!("foo");
        assert!(jsonschema::is_valid(&schema, &instance));
        assert!(jsonschema::validate(&schema, &instance).is_ok());
    }
    println!("Allocated after validation: {}KB", ALLOCATOR.allocated()/1024);
```
```
Allocated before validation: 0KB
Allocated after validation: 4184KB
```

Creating a more complex schema seems to increase the memory held after the validation scope:
```
println!("Allocated before validation: {}KB", ALLOCATOR.allocated()/1024);
    {
        let schema = json!({"type":"object", 
                            "properties": { 
                                "innerThing" : { 
                                    "type" :"string", 
                                    "maxLength": 5, 
                                    "minLength": 1
                                }, 
                                "anotherThing" : {
                                    "type":"string", 
                                    "maxLength":10, 
                                    "minLength": 1
                                }, 
                                "arrayThing" : {
                                    "type": "array", 
                                    "items": {
                                        "type" : "string"
                                    }
                                } 
                            }, 
                            "required": ["innerThing"]
                        });
        let instance = json!("foo");
        _ = jsonschema::is_valid(&schema, &instance);
        _ = jsonschema::validate(&schema, &instance).is_ok();
    }
[...]

Issue #641, opened 2024-11-22
Execution model: Register-based Virtual Machine


Stranger6667:
### Summary

This proposal is about reworking the JSON Schema compiler, implementing a virtual machine for validating JSON Schemas, and integrating multiple backends into the compilation pipeline.

The main goal is to change the execution model to a more performance-oriented one. Bytecode execution would be a much faster approach than the current tree-walk interpreter, and it is possible to generate Rust code when the schema is known statically.

### Motivation

The current implementation uses super instructions to optimize validation by combining multiple JSON Schema keywords into single operations, minimizing jumps and execution overhead. 

While decently effective, this approach is bound by the limitations of direct interpretation. A virtual machine executing bytecode could significantly improve performance by:

1. Better spatial locality as instructions are packed in a single allocation.
2. Avoiding lazy compilation as the validation state becomes smaller (instruction pointer + a few iterators + stack of call frames) and easier to handle.
3. Precompilation of schemas into efficient Rust code via procedural macros / `build.rs`.
4. Allowing schemas to be serialized into bytecode, stored, and reloaded for reuse, enabling VM execution directly from a block of static memory.

### Implementation Overview

1. Rework the compilation pipeline so it generates JSON Schema Intermediate Representation (JSIR) first (POC done)
2. Add JSIR-specific optimization passes (canonicalization like removing redundant combinators like single-value `allOf` / `anyOf` + `$ref` inlining, loop unrolling, etc). Requires a bit of extra case to properly store location metadata
3. Compilation backend for VM + optimization passes (superinstructions, redundant jumps elimination, etc).
4. Compilation backend for Rust code. 

Probably (3) and (4) may require their own IR format to simplify codegen, i.e. I can imagine  instructions like `IF_UNPACK_NUMBER` that will generate code like 

```rust
if let Value::Number(number) = value {
} else {
}
```

Right now I've implemented direct bytecode compilation + interpretation for some JSON Schema subset, but plan to split this process into multiple phases.

PR #749, opened 2025-05-22
refactor: Use index vec to store the schema nodes

Stranger6667:
This is yet another attempt to improve the schema tree layout without going fully into building a compiler & interpreter, as noted in #641 (which will require way more effort).

The core goal is to prevent excessive memory consumption by recursive references, which are evaluated lazily and keep expanding on each deeper recursion cycle (#675).

Solving the above issue should also remove the need to store `Registry` inside `Validator`, and, in principle, it should be possible to pass `Registry` by reference. It will unblock a better API for `jsonschema` where the user won't need to pass the `Registry` by value (likely cloning it).

Inspired by changes I made in the `css-inline` crate a few years ago.


lucab:
I was looking a bit more into the memory usage of this library, and I have an additional sidenote that I'll leave here.
Other than the well-known lazy-statics, it looks like there are some larger memory leak coming from the `$ref` validator here:
https://github.com/Stranger6667/jsonschema/blob/958b430eedc2a6c96810bba2a8b3bad08fa61217/crates/jsonschema/src/keywords/ref_.rs#L124-L142

I think that the current recursive-validators manage to build a cyclic family of `Arc` (or `Rc`?) via the intertwined `Context`, `Registry`, `Resolver` and the validators.
I naively tried to break this by demoting `LazyRefValidator.registry` to a `Weak`, but that didn't work at runtime.

Stranger6667:
@lucab 

Thank you for taking a deep look into this! The presence of registry, ctx, etc, is surely an awful hack. In this PR, the static references are resolved at the compilation step, so there is no need to keep those structures.

I am not sure about `$dynamicRef` 100% yet, but it should also be possible to avoid this "compilation at runtime" approach too. I.e., store all possible jump locations and decide on the exact one at runtime.

Right now, static refs work in the new compiler, so adding `$dynamicRef` is the next step on my TODO list.

As a side note, another factor contributing to memory usage could be the presence of many separate allocations (every node is boxed). I believe that, depending on the used allocator, memory fragmentation may also impact the final result.

Gradually, I want to pack things more (e.g. intern `String` for properties, etc), and I think it will also improve the situation with runtime performance & memory usage.

Stranger6667:
Reworked the recursive references recently, so the memory consumption issue is solved. I'll continue to make more improvements to the compiler later on, separately.

PR #750, opened 2025-05-24
fix: multiple_of_float

lovebug356:
It appears that multiple issues have been reported regarding the validation of multipleOf for floating-point values. I have also identified a problem with the initial implementation of the remainder calculation: on my system, several (newly added) tests fail. Although the code implies that the results should fall within an EPSILON tolerance, the deviations are substantial. When BigFraction is used instead, all tests pass successfully.

value: 12.2 multipleOf:0.1 remainder:0.9999999999999858
value: 47.498 multipleOf:0.0001 remainder:0.9999999999417923
value: 3.1254 multipleOf:0.0001 remainder:0.999999999996362


Stranger6667:
Thank you so much! I'll update the changelog and will make a new release soon

PR #757, opened 2025-06-04
perf(referencing): avoid cloning static Values for meta-schemas

lucab:
This tweaks the meta-schemas logic related to building lazy-static registries, in order to avoid cloning all the inner JSON Values which already have `'static` lifetimes.


Stranger6667:
Oh, nice! Thank you :)

PR #758, opened 2025-06-04
feat: added support for string enums in object keys #453

### LINKED ISSUE #453: Use string enums as dictionary keys
Currently, a string enum cannot be used as the key of an object.

```py
from enum import Enum, StrEnum
from jsonschema_rs import is_valid

schema = {
    "$schema": "http://json-schema.org/draft-07/schema#",
    "additionalProperties": False,
    "properties": {
        "foo": {"type": "number"},
        "baz": {"type": "number"},
    },
    "required": ["foo", "baz"],
    "type": "object",
}


class MyEnum(str, Enum):
    foo = "foo"
    bar = "baz"


# this works
print(is_valid(schema, {"foo": 1, "baz": 2}))

# this works too
print(is_valid(schema, {MyEnum.foo.value: 1, MyEnum.bar.value: 2}))

# this raises a ValueError: Dict key must be str. Got 'MyEnum'
print(is_valid(schema, {MyEnum.foo: 1, MyEnum.bar: 2}))
```

To me, it seems a valid use case of a string enum, since:
- enums in values are supported by the library
- string enums should always have a string `value` property. 

I've tried with the 3.11 `StrEnum` type, that guarantees that `str(MyEnum.foo) == "foo"`, but it doesn't work either (probably because I saw an explicity check on the type of the key to be exactly str).

Is it feasible to:
- for classes that subclass `(str, Enum)` to use their `value` property
- for classes that subclass `StrEnum` to use their `__str__` representation?

DavideCanton:
Resolves #453

In the end I chose to always use the `value` property in both old and new style enums

### DIFF
--- a/crates/jsonschema-py/CHANGELOG.md
+++ b/crates/jsonschema-py/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added support for old and new style string enums in object keys.
+
 ### Changed
 
 - Update `pyo3` to `0.25`.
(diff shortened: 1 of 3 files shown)


Stranger6667 (approved):

Stranger6667:
Awesome! Thank you :)

PR #763, opened 2025-06-13
refactor: shard meta-schemas per draft

lucab:
This introduces some new (internal-only and lazy) static entities in order to shard meta-schemas according to the draft that they belong to. Sharded meta-schemas allow to reduce the amount of registry resources that need to be constructed and statically cached.


Stranger6667:
Amazing! Thank you so much! Do you have some numbers on the memory usage change?

lucab:
I have some numbers from a private application, but unfortunately I can't share the code nor the JSON schema/docs. 

As a single data-point, performing validation of a single doc and a single schema under draft-202012, this seems to be saving ~300kB of lazy-cached memory. I'm checking via https://github.com/KDE/heaptrack.

I'm honestly a bit disappointed as I was hoping for more, but at this point I think most of the memory usage is coming from the lazy/recursive `$ref` handling.

Stranger6667:
Aha, I see. 

300kb is not nothing, and the change seems to be relatively straightforward :)

Yesterday, I got at least some stuff working in #749, and I believe that approach should decrease the memory usage significantly, as it allows using an id to refer to a subschema, instead of building that lazy validator, which currently is implemented for `$ref`

lucab:
Thanks, I'll try to have a look at #749 if I manage to (I'm currently traveling).
If there are no major conflicts and this PR seems overall fine, I think you can land this first then rebase your ongoing work.

Stranger6667:
Thank you!

Issue #649, opened 2024-12-22
Add `context` field to `ValidationError`, to improve oneOf and anyOf validation errors


jpmckinney:
In the case where no subschema is valid, the ValidationError instances in python-jsonschema set a `context` attribute for allOf and anyOf validation errors as [documented](https://python-jsonschema.readthedocs.io/en/stable/errors/) and in the [code](https://github.com/python-jsonschema/jsonschema/blob/93e0caa5752947ec77333da81a634afe41a022ed/jsonschema/_keywords.py#L338-L364).

This `context` attribute contains all the validation errors from the subschemas. This is very useful, because the default error message is not sufficiently informative: "[frequently a very large JSON blob] is not valid under any of the schemas listed in the 'oneOf' keyword".

Instead, an application could report the specific errors under each of the subschemas (or, it could have some way to determine which subschema was most relevant, and only report its errors). It might then end up reporting a single, specific validation error within the subschema, which is much more actionable. For example, "'name' is not a string".

Adding this sort of `context` attribute (I have no opinion on naming) would allow applications to improve the error message.

Stranger6667:
This is super useful indeed, and I would be happy to have a mechanism to improve error messages for such cases (personally I saw a lot of opaque error messages of this sort while validating Open API with jsonschema). 

The biggest challenge is avoiding a performance hit by collecting all the related errors. The first idea that comes to mind is to re-validate on demand and enable this "collection" logic, so this way the users who don't need this context won't be affected. Not sure though

jpmckinney:
Could there be an argument passed to validator_for (and other initializers) that either:

1. gets passed to OneOfValidator via the ctx argument, such that it uses iter_errors instead of is_valid when checking subschemas (this means some extra if-statements in the oneOf validator)
2. gets passed to get_for_draft, such that a different oneOf validator is chosen, that uses iter_errors (no extra instructions from if-statements, but probably some code duplication)

orbitz:
Hello!  I recently incorporated this into our product to give users better feedback on configuration mistakes.  Better oneOf support would be really helpful because we use it a lot to maintain backwards compatibility in our configuration.

In our specific case we are less concerned about performance for this because we only do the JSON schema validation if reading the config file fails. As in, we have a translation layer that succeeds or fails and only if it fails do we validate it against the schema to get a better error message.


Let me know if there is any way to help testing or otherwise support this ticket.

Thank you!

orbitz:
[...]

PR #772, opened 2025-07-21
Add `context` field to `anyOf` and `oneOf` validation error

kprzybyla:
Fixes https://github.com/Stranger6667/jsonschema/issues/649

Hi, thank you for this awesome library! We are currently using the Python `jsonschema` library, and we want to switch to the `jsonschema-rs`, but we are blocked by one missing feature, which is the context for the `anyOf` and `oneOf` validation. We are currently using the `context` present in the `anyOf` and `oneOf` to provide better errors for our end-users.

Now I do understand that collecting those additional errors for the context of `anyOf` and `oneOf` can have an impact on the performance, so I went with the route of minimizing additional allocations and code branches for the happy path by simply collecting those errors only in case of an error during the validation. This does mean that the validation is executed again in order to collect errors, but it also avoids calling `.iter_errors()` instead of `.is_valid()` for the happy path, during which we actually want the most performance. I think this additional overhead should be fine for the unhappy path, as it should be hit rarely in the production systems, and it should greatly improve the user experience when the error happens.

If you have a different idea how to solve this, I'm willing to work on this, as this is essential for us to switch to `jsonschema-rs`. The only thing right now that I don't like with this approach is that those errors are eventually converted to `'static` objects, which live for the whole duration of the program because `ValidationErrorKind` does not have a lifetime. This could lead to leaking memory every time we hit the validation error for the `anyOf` or `oneOf`. I started tinkering with the idea of adding lifetime to `ValidationErrorKind`, but it does not seem to be so trivial. Please let me know what you think about that.

Also, during the development of this, I found one small bug in the Python bindings (as I made the same mistake myself) and added a simple fix for it in a separate commit, see below:
- https://github.com/PyO3/pyo3/issues/3747
- https://pyo3.rs/v0.25.1/class#complex-enums.

> WARNING: Py::new and .into_pyobject are currently inconsistent. Note how the constructed value is not an instance of the specific variant. For this reason, constructing values is only recommended using .into_pyobject.

### DIFF
--- a/crates/jsonschema/src/error.rs
+++ b/crates/jsonschema/src/error.rs
@@ -781,13 +781,13 @@ fn write_quoted_list(f: &mut Formatter<'_>, items: &[impl fmt::Display]) -> fmt:
     let mut iter = items.iter();
     if let Some(item) = iter.next() {
         f.write_char('\'')?;
-        write!(f, "{}", item)?;
+        write!(f, "{item}")?;
         f.write_char('\'')?;
     }
     for item in iter {
         f.write_str(", ")?;
         f.write_char('\'')?;
-        write!(f, "{}", item)?;
+        write!(f, "{item}")?;
         f.write_char('\'')?;
     }
     Ok(())
@@ -817,11 +817,11 @@ impl fmt::Display for ValidationError<'_> {
                 let mut iter = array.iter().skip(*limit);
 
                 if let Some(item) = iter.next() {
-                    write!(f, "{}", item)?;
+                    write!(f, "{item}")?;
                 }
                 for item in iter {
                     f.write_str(", ")?;
-                    write!(f, "{}", item)?;
+                    write!(f, "{item}")?;
                 }
 
                 write_unexpected_suffix(f, array.len() - limit)
@@ -847,7 +847,7 @@ impl fmt::Display for ValidationError<'_> {
                 self.instance
             ),
             ValidationErrorKind::Constant { expected_value } => {
-                write!(f, "{} was expected", expected_value)
+                write!(f, "{expected_value} was expected")
             }
             ValidationErrorKind::ContentEncoding { content_encoding } => {
                 write!(
@@ -943,7 +943,7 @@ impl fmt::Display for ValidationError<'_> {
             }
             ValidationErrorKind::PropertyNames { error } => error.fmt(f),
             ValidationErrorKind::Required { property } => {
-                write!(f, "{} is a required property", property)
+                write!(f, "{property} is a required property")
             }
             ValidationErrorKind::MultipleOf { multiple_of } => {
                 write!(f, "{} is not a multiple of {}", self.instance, multiple_of)
@@ -971,13 +971,13 @@ impl fmt::Display for ValidationError<'_> {
                 let mut iter = types.iter();
                 if let Some(t) = iter.next() {
                     f.write_char('"')?;
-                    write!(f, "{}", t)?;
+                    write!(f, "{t}")?;
                     f.write_char('"')?;
                 }
                 for t in iter {
                     f.write_str(", ")?;
                     f.write_char('"')?;
-                    write!(f, "{}", t)?;
+                    write!(f, "{t}")?;
                     f.write_char('"')?;
                 }
                 Ok(())
@@ -1026,7 +1026,7 @@ impl fmt::Display for MaskedValidationError<'_, '_, '_> {
                 self.placeholder
             ),
             ValidationErrorKind::Constant { expected_value } => {
-                write!(f, "{} was expected", expected_value)
+                write!(f, "{expected_value} was expected")
             }
             ValidationErrorKind::ContentEncoding { content_encoding } => {
                 write!(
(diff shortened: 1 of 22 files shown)


Stranger6667:
Hi @kprzybyla 

thank you for bringing this up! I am on vacation right now so I will review the PR a bit later, but I am happy that this issue moves forward :)

crates/jsonschema/src/error.rs
```diff
@@ -98,7 +98,7 @@ pub enum ValidationErrorKind {
     /// Unexpected properties.
     AdditionalProperties { unexpected: Vec<String> },
     /// The input value is not valid under any of the schemas listed in the 'anyOf' keyword.
-    AnyOf,
+    AnyOf { context: Vec<(usize, Vec<ValidationError<'static>>)> },
```
Stranger6667: As it is a breaking change anyway, should we add a lifetime to `ValidationErrorKind` instead of keeping `'static` here?

crates/jsonschema/src/keywords/any_of.rs
```diff
@@ -70,6 +80,16 @@ impl Validate for AnyOfValidator {
...
+                self.schemas
+                    .iter()
+                    .enumerate()
+                    .map(|(schema_id, schema)| (
+                        schema_id,
+                        schema
+                            .iter_errors(instance, location)
+                            .collect()
+                    ))
+                    .collect(),
```
Stranger6667: Do we actually need `schema_id` as it could be inferred from the vector of results here? i.e., the caller can call `enumerate` on the context later on if needed and get the schema_id. 

So the relevant enum variant could be `AnyOf { context: Vec<Vec<ValidationError<'static>>> },`

Stranger6667:
Re: leaking memory. Could you please elaborate? `ValidationError::to_owned` converts the error instance into an owned value, so it should not be any different from any other owned value on the Rust side and could be deallocated the same way as any other value (e.g., `String` instances we already have inside bindings). I.e. the static lifetime we have there is merely a way to pass owned values to the Python side - there is no `Box::leak` or similar

kprzybyla:
> Hi @kprzybyla
> 
> thank you for bringing this up! I am on vacation right now so I will review the PR a bit later, but I am happy that this issue moves forward :)

Great, thanks! I will fix all pipeline issues in the meantime :)

> Re: leaking memory. Could you please elaborate? `ValidationError::to_owned` converts the error instance into an owned value, so it should not be any different from any other owned value on the Rust side and could be deallocated the same way as any other value (e.g., `String` instances we already have inside bindings). I.e. the static lifetime we have there is merely a way to pass owned values to the Python side - there is no `Box::leak` or similar

Ok, you are right, sorry about that. I'm not super fluent in Rust, so it confused me a little bit at first but I get it now :)

crates/jsonschema/src/keywords/any_of.rs
kprzybyla: Yes, that's a good idea, the index of the element will match the `schema_id`, so we are duplicating information here.

crates/jsonschema/src/error.rs
kprzybyla: I will look into this further, as I already experimented with adding lifetime to `ValidationErrorKind`, but unfortunately, this will most likely require adding lifetime annotations in a lot of places where `ValidationErrorKind` is used.

Stranger6667: Aha, I think it is not critical right now, so it could come later on as an optimization

Stranger6667: And also I am not sure if it is possible to do at all with non static lifetime. Because the errors are created there in the error branch, so they should be stored somewhere, and the owned version of Cow inside ValidationError looks like the only possible way to do so

Stranger6667:
Looks great to me!

Only changelog entries left (one for the Rust crate, one for the Python bindings).

The failing lint is not directly connected to this PR, so it is up to you if you'd like to fix it - or I can push a fix afterward

kprzybyla:
@Stranger6667 Great, I will push the changes in the changelogs in a second

I can fix the lint issues, I just noticed that it complained in other places too.

Stranger6667:
Cool! There are actually more lint issues, but clippy is not properly configured right now - fixing only the basic ones that clippy shows right now is fine.

kprzybyla:
Ok, all should be good now with the lint, and I added entries in the changelogs.

Stranger6667:
Amazing! Looks great to me! Thank you for your contribution :) I'll take a look at the current docs and maybe update them with this new feature during the next few days

Also, maybe @jpmckinney and @orbitz have some feedback - I'd be happy to learn what you think and if these changes are sufficient for the original use cases noted in #649. In any event, we can build more things in subsequent releases

Stranger6667:
Checking the original issue I found that it also mentions allOf - it could also be useful to have that context for it

kprzybyla:
@Stranger6667 I had a draft with adding context to allOf but I scrapped it since a couple of jsonschema spec tests were failling due to those changes. But I can revisit it again. In our code base we did not look into `allOf` context and the `.iter_errors()` already returns all errors from all schemas so that's why I skipped it. But `.validate()` only returns first error so maybe someone would prefer to also have a context that contains all other errors.

Stranger6667:
@kprzybyla 

Okay, we can always extend it later :) I am back from vacation and will take a look at this PR deeper soon, and I think we are good to merge it!

Stranger6667:
Released in 0.31.0

PR #903, opened 2025-12-02
feat: email validation option

### LINKED ISSUE #894: email validation options?
"missing@domain" validates as a valid email, which it technically is per the spec.

https://github.com/johnstonskj/rust-email_address/issues/38

IRL however, should it be?

Should email validation options be exposed so the jsonschema crate user can adjust the email validation behavior?

https://docs.rs/email_address/latest/email_address/struct.Options.html

jqnatividad:
resolves #894

### DIFF
--- a/crates/jsonschema/src/keywords/format.rs
+++ b/crates/jsonschema/src/keywords/format.rs
@@ -839,7 +839,7 @@ struct IdnEmailValidator {
 impl IdnEmailValidator {
     pub(crate) fn compile<'a>(ctx: &compiler::Context) -> CompilationResult<'a> {
         let location = ctx.location().join("format");
-        let email_options = ctx.config().email_options().cloned();
+        let email_options = ctx.config().email_options().copied();
         Ok(Box::new(IdnEmailValidator {
             location,
             email_options,
@@ -1008,7 +1008,7 @@ mod tests {
     use serde_json::json;
     use test_case::test_case;
 
-    use crate::tests_util;
+    use crate::{tests_util, EmailOptions};
 
     use super::*;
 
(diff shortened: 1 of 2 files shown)


crates/jsonschema/src/keywords/format.rs
```diff
@@ -1205,4 +1305,211 @@ mod tests {
...
+            .expect("Schema should compile");
+
+        assert!(!validator.is_valid(&json!("user@example.com")));
+        assert!(validator.is_valid(&json!("user@sub.example.com")));
+    }
+
+    #[test]
+    fn email_options_domain_literal() {
+        use crate::EmailOptions;
+        use email_address::Options as EmailAddressOptions;
```
Stranger6667: I think this one could be extracted to module level `use` to avoid repeating it in every test function

crates/jsonschema/src/options.rs
```diff
@@ -829,6 +860,56 @@ impl Default for PatternEngineOptions {
...
+/// let validator = jsonschema::options()
+///     .with_email_options(EmailOptions::default())
+///     .should_validate_formats(true)
+///     .build(&schema)
+///     .expect("A valid schema");
+/// ```
+#[derive(Debug, Clone)]
+pub struct EmailOptions {
+    pub(crate) inner: EmailAddressOptions,
+}
```
Stranger6667, line 897: I think it would be nice to have an API to configure the actual options, otherwise it feels somewhat limited for cases when the user wants to customize it. I.e. the current option would be to have that `EmailAddressOptions` and configure it first, so the user will need to add that crate to dependencies too + create that instance first - I'd like to avoid that.

crates/jsonschema/src/options.rs
```diff
@@ -829,6 +860,56 @@ impl Default for PatternEngineOptions {
...
+        Self {
+            inner: EmailAddressOptions::default(),
+        }
+    }
+
+    /// Create email validation options from `email_address::Options`.
+    ///
+    /// This allows full access to all options provided by the `email_address` crate.
+    #[must_use]
+    pub fn from_options(options: EmailAddressOptions) -> Self {
```
Stranger6667: `From` would be nice

PR #988, opened 2026-02-06
Add yaml support to jsonschema-cli

### LINKED ISSUE #985: Feature request: be able to run a schema against a yaml instance
Hi, thanks for the impressive work.

I would like to use the cli to validate `yaml` files against a `json-schema`:

`jsonschema-cli schema.json -i instance.yaml -i or-instance.yml`

It could be behind a `feature-flag`

kod-kristoff:
This pull request adds YAML support to `jsonschema-cli`.

This is done by checking if the extension of the given file is `yaml` or `yml`, an explicit alternative would be to add the flag `-y` for YAML-files. But that would be a much larger change.

For the yaml reading part, I added the library `serde-saphyr`.

Error message: `Error: failed to read JSON from bad.json: expected `:` at line 1 column 7`

Fixes #985

### DIFF
--- a/crates/jsonschema-cli/src/main.rs
+++ b/crates/jsonschema-cli/src/main.rs
@@ -191,6 +191,67 @@ fn read_json(
     Ok(serde_json::from_reader(reader))
 }
 
+#[derive(Debug)]
+enum ReadJsonOrYamlError {
+    Json {
+        file: PathBuf,
+        err: serde_json::Error,
+    },
+    Yaml {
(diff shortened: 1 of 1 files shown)


Stranger6667:
Thanks for opening this! I like the approach :) I was thinking about releasing 0.42 today's evening. Hence I am thinking about merging this PR as is and I'll follow up with test fixes and extra tests for CLI + changelog entry. Would you be fine with that?

PR #996, opened 2026-02-13
test: sort results so that tests are stable

kod-kristoff:
This PR makes the unit tests for `jsonschema` stable. 

This is done be sorting the results where it is necessary.


Stranger6667:
Thanks!

PR #1055, opened 2026-03-01
feat: `canonical::json::to_string` for canonical JSON serialization

Stranger6667:


crates/jsonschema/src/canonical.rs
```diff
@@ -0,0 +1,430 @@
...
+    };
+
+    estimated.clamp(MIN_CAPACITY, MAX_PREALLOC)
+}
+
+/// A formatter that emits integer-valued floats as integers.
+#[derive(Default)]
+struct CanonicalFormatter {
+    default: CompactFormatter,
+}
```
cschramm, line 115: This requires `serde_json` `1.0.143` while the workspace and thus the crate specifies just `1` (which was probably too lax already)

PR #1075, opened 2026-03-16
feat: Validate with JSON pointer

### LINKED ISSUE #432: Validating individual definitions
I'm trying to validate JSON snippets against the DAP JSON schema: https://microsoft.github.io/debug-adapter-protocol/debugAdapterProtocol.json

Each type of message is a separate definition. I can extract and compile the relevant definition, but references are not resolved. I think I need to use the `with_document` function, but I'm not sure what the ID is supposed to be. 

```rs
fn main() {
    let schema = include_str!("../debugAdapterProtocol.json");
    let schema: Value = serde_json::from_str(schema).unwrap();

    // Get the specific definition from the schema
    let next_request_schema = schema.get("definitions").unwrap().get("NextRequest").unwrap().clone();
    
    // Compile this specific schema
    let compiled = JSONSchema::options()
        .should_validate_formats(true)
        .should_ignore_unknown_formats(false)
        .with_document("#".to_string(), schema)
        .compile(&next_request_schema)
        .unwrap();

    let instance = json!({
        "seq": 153,
        "type": "request",
        "command": "next",
        "arguments": {
        "threadId": 3
        }
    });
    let result = compiled.validate(&instance);
    if let Err(errors) = result {
        for error in errors {
            println!("Validation error: {}", error);
            println!("Instance path: {}", error.instance_path);
        }
    }
}
```

This results in

```
Validation error: Invalid reference: json-schema:///#/definitions/Request
Instance path:
Validation error: Invalid reference: json-schema:///#/definitions/NextArguments
Instance path:
```

I tried an empty string as well, but that didn't work either. Any clues?

yahelnachum:
Some documents, such as OpenAPI, do not contain a referenced root schema, but do contain multiple schemas that users would like to validate against. The current solution to this is to create a resource of this document and point into the resource with a small additional root schema. Instead, users should be able to compile the document and at a later time specify a subschema, using a JSON pointer, at execution time.

Summary of code changes:
- Added an option to collect all <JSON pointer, subschema> pairs for compile-time
- Added a function to validate with a JSON pointer for execution-time

Resolves #432


Stranger6667:
Sorry for the delay! I am going to review it next week - need to rework some parts of the registry-related workflows first.

yahelnachum:
No problem, let me know what you think!

yahelnachum:
@Stranger6667 have you had a chance to look at this? Thanks again for your time!

Stranger6667:
@yahelnachum Sorry for the delay!

My main concern is that the `Validator` struct gains extra responsibility and transitions from "single validator" into "collection of validators" - my preference would be to have a separate "collection" type for this, that will have a small API surface on its own and will expose the current `Validator` from it. 

Right now there are new error paths (when `subschemas` is `None`) that could be avoided if we organize structures differently (i.e. the user always constructs "collection" explicitly and there is nothing to check).

For example, it could be a new "build"-like method on `ValidationOptions` that will create this new structure (`ValidatorMap` so it is easier to maybe have some internal changes to map itself in the future / change the map type?), then each `Validator` can be accessed via a map-like API:

```rust
let map = jsonschema::options().build_map(&schema)?;
let user_validator = map["/components/schemas/User"];  # or map.get("...")
user_validator.is_valid(...);
```

I think this way, the API expansion scope is smaller and the feature itself is harder to misuse

yahelnachum:
@Stranger6667 thanks for the feedback! Yes, I like that that the method you proposed would prevent these weird runtime error scenarios I had to return on 👍. I think I can make these type of changes hopefully in the next 2 weeks. I can ping you when I have completed them.

Stranger6667:
I'll push a few changes here and hopefully finish the implementation

PR #1109, opened 2026-04-11
Use PyDict_Size instead of raw dict layout access

smarcd:
This replaces the direct `PyDictObject.ma_used` layout read with `PyDict_Size`.

Why:
- the current code relies on CPython dict object layout details
- `PyDict_Size` expresses the same intent through the public C-API
- this is more portable across Python runtimes and ffi backends while preserving behavior on CPython

The change is intentionally narrow: it only updates `dict_len()` to use `PyDict_Size` and handles the error case conservatively.


Stranger6667:
Thanks!

PR #1159, opened 2026-06-21
feat: Support `IntoRegistryResource` for `Arc<Value>`

### LINKED ISSUE #1129: Implement `IntoRegistryResource` for `Arc<Value>`
It would sometimes be nice to be able to keep a `Value` in a `Registry` without consuming it entirely, such as when the `Value` is large and expensive to clone. It seems like `Registry` only needs a shared reference: would it be possible to implement `IntoRegistryResource` for `Arc<Value>`?

A way to get a borrow of the `Value` from a `Registry` would also work, but I'm not sure how feasible that is.

Stranger6667:
Resolves #1129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


tgross35:
Thank you for adding this!

PR #1440, opened 2026-08-07
perf: Avoid repeated definitions traversal

Stranger6667:


Stranger6667:

PR #1441, opened 2026-08-07
perf: 13% faster canonicalization of a non-dynamic OpenAPI document, which no longer scans every schema object for dynamic references

Stranger6667:


Stranger6667:

PR #1442, opened 2026-08-07
perf: Faster canonicalization of schemas with local `$defs` references, which avoid decoding unescaped definition names

Stranger6667:


Stranger6667:

PR #1443, opened 2026-08-07
perf: 5% faster emission of canonical schemas, which no longer formats object keys while copying them

Stranger6667:


Stranger6667:
@codspeedbot Is it a real improvement?

PR #1445, opened 2026-08-07
perf: 2% faster emission of canonical schemas, which no longer formats copied string values

Stranger6667:


Stranger6667:
@codspeedbot is it a real perf improvement even if it is under the threshold?

PR #1446, opened 2026-08-07
perf: 2% faster canonicalization of schemas that assert known string formats

Stranger6667:


Stranger6667:
@codspeedbot is it a real perf improvement even if it is under the threshold?

Stranger6667:
@codspeedbot I didn't ask you to commit anything

PR #1457, opened 2026-08-08
perf: Adjust intersection

Stranger6667:


Stranger6667:
@codspeedbot - does this change affect anything, even below the threshold?

Stranger6667:
@codspeedbot - re-check the performance metrics for this change

PR #1569, opened 2026-09-04
feat: `PreparedDocument::unsatisfiable_pointers`, listing every subschema no value satisfies

Stranger6667:


Stranger6667:
@codspeedbot explain this regression

PR #1574, opened 2026-09-05
fix: Fused array shape checking the length bound when a custom keyword overrides minItems/maxItems

sagikazarmark:
Since 0.49, `type: "array"` + `items` + `minItems`/`maxItems` is fused into a
single `ArrayShapeValidator`. The built-in `minItems`/`maxItems` compilers know
about this and return `None` when the fusion applies, so the bound is checked
once.

The compiler dispatches to custom keyword factories before the built-in ones,
but `array_shape_fusion` never checks whether a custom factory is registered
for `minItems` or `maxItems`. When one is, the schema ends up with two
validators for the same bound: the fused built-in one and the custom one.

This shows up as duplicate errors for one violation, and it defeats the point
of overriding the keyword. For example, a custom `minItems` that keeps bounds
past `u64` exact still gets a second, saturated `u64::MAX` check from the fused
validator.

Fix: `array_shape_fusion` bails out when a custom factory is registered for
`minItems` or `maxItems`. `type` and `items` then compile through the regular
path, and the custom keyword is the only thing checking the bound.

Includes a regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Fixed
 
+- `type: array` with `items` and `minItems`/`maxItems` checking the built-in length bound even when a custom keyword overrides `minItems` or `maxItems`, so the bound was evaluated twice.
 - `#[jsonschema::validator]` generated code that named `serde_json` directly, so it needed the consumer crate to depend on `serde_json` too.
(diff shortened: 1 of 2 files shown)


Stranger6667 (approved):
Thank you for fixing this!

Stranger6667:
there are a few more similar cases, I'll fix them separately

sagikazarmark:
Thanks @Stranger6667!

I was only affected by this, so only fixed this one, but at least it shed some light on this type of issue. 🙂 

Thanks for maintaining this library!

PR #1577, opened 2026-09-05
perf: Misc performance improvements

Stranger6667:


Stranger6667:
@codspeedbot explain why performance improved

PR #1600, opened 2026-09-09
perf: Canonicalizing several subschemas of one prepared document compiles each pattern once for the document instead of once per call

Stranger6667:


Stranger6667:
@codspeedbot explain performance effects

PR #1607, opened 2026-09-11
perf: Canonicalizing several subschemas of one prepared document parses each definition body once for the document instead of once per call

Stranger6667:


Stranger6667:
@codspeedbot fix this regression

PR #1613, opened 2026-09-15
ci: add Windows ARM64 support for Rust and Python

ndabas:
A couple of things to note:
- CPython has releases for Windows ARM64 starting from 3.11, and PyPy does not have any, so the matrices reflect that.
- I swapped out `uv python install` for `actions/setup-python`. `uv` does not install native Python by default on Windows ARM64 (installing an x86-64 release under emulation instead), and has no free-threaded builds available for that platform.

All of the workflows I have edited are passing on my fork.


Stranger6667:
Thank you!

PR #1625, opened 2026-09-23
feat: add `should_deny_unknown_keywords` validation option

luca-iachini:
Keywords the draft does not define are kept as annotations and otherwise ignored, so a typo like "reqired" silently yields a schema that accepts everything. This PR adds `should_deny_unknown_keywords(bool)`, `false` by default, which makes building fail instead:

```rust
let error = jsonschema::options()
    .should_deny_unknown_keywords(true)
    .build(&json!({"reqired": ["a"]}))
    .expect_err("Should reject unknown keyword");

assert_eq!(
    error.to_string(),
    "Unknown keyword: 'reqired'. Adjust configuration to ignore unrecognized keywords"
);
```

The check walks the document before compilation rather than sitting inside it, since compilation follows references and an unused `$defs` entry would never be reached. 
Annotation keywords are allowed for the drafts that define them, as are keywords registered via `with_keyword` and an embedded resource's own keywords. It leaves alone `const` bodies, `enum` members and property names, resources pulled from a registry, and the `additionalItems` and `dependencies` subschemas under Draft 2020-12. This required making `referencing::Draft::walk_children` public; it already existed for registry building, and the walk needs the keyword path for error locations.


Stranger6667:
Generally I'd prefer discussing ideas via issues first, but I like this use case. However, I am skeptical of the current implementation as it adds yet another traversal. Having slightly adjusted metaschemas would avoid most of the logic changes + duplication between the actual schemas + rust code (i.e. I'd prefer to have some duplication in metaschemas):

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "urn:jsonschema-rs:strict-meta:draft2020-12",
  "$dynamicAnchor": "meta",
  "$ref": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "$anchor": true, "$comment": true, "$defs": true, "$dynamicAnchor": true, "$dynamicRef": true,
    "$id": true, "$recursiveAnchor": true, "$recursiveRef": true, "$ref": true, "$schema": true,
    "$vocabulary": true, "additionalProperties": true, "allOf": true, "anyOf": true, "const": true,
    "contains": true, "contentEncoding": true, "contentMediaType": true, "contentSchema": true,
    "default": true, "definitions": true, "dependencies": true, "dependentRequired": true,
    "dependentSchemas": true, "deprecated": true, "description": true, "else": true, "enum": true,
    "examples": true, "exclusiveMaximum": true, "exclusiveMinimum": true, "format": true, "if": true,
    "items": true, "maxContains": true, "maxItems": true, "maxLength": true, "maxProperties": true,
    "maximum": true, "minContains": true, "minItems": true, "minLength": true, "minProperties": true,
    "minimum": true, "multipleOf": true, "not": true, "oneOf": true, "pattern": true,
    "patternProperties": true, "prefixItems": true, "properties": true, "propertyNames": true,
    "readOnly": true, "required": true, "then": true, "title": true, "type": true,
    "unevaluatedItems": true, "unevaluatedProperties": true, "uniqueItems": true, "writeOnly": true
  },
  "additionalProperties": false
}
```

or

```json
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "$id": "http://json-schema.org/draft-07/schema#",
     "title": "Core schema meta-schema",
     "definitions": { ... },
     "type": ["object", "boolean"],
     "properties": { ... 46 keywords, unchanged ... },
+    "additionalProperties": false,
     "default": true
 }
```

then:

```rust
#[jsonschema_macros::validator(
    path = "metaschemas/strict/draft2020-12.json",
    draft = Draft202012,
    resources = {
        "https://json-schema.org/draft/2020-12/schema" => { path = "metaschemas/draft2020-12/schema.json" },
        "https://json-schema.org/draft/2020-12/meta/core" => { path = "metaschemas/draft2020-12/meta/core.json" },
        // ... the other six vocabularies, as today
    }
)]
struct StrictMetaDraft202012;
```

so it will return `AdditionalProperties` from `iter_errors` which can be filtered against custom keywords + error message could be adjusted there too. The runtime version is similar (without the `macros` feature).

luca-iachini:
> Generally I'd prefer discussing ideas via issues first, but I like this use case.

Sorry, I missed `CONTRIBUTING.md` line. I should have opened an issue before submitting the PR. I’m happy to close this and move the discussion to an issue if you prefer.

> However, I am skeptical of the current implementation as it adds yet another traversal. Having slightly adjusted metaschemas would avoid most of the logic changes + duplication between the actual schemas + rust code (i.e. I'd prefer to have some duplication in metaschemas):

I think the two approaches answer different questions.

A strict meta-schema is something users can already use today without any changes to the library: they can write the meta-schema themselves and validate their schema against it. What I’m proposing is slightly different: making this a behavior of the compiler, so that it can report when it encounters unknown keywords. Also with your proposal requires the library to maintain custom meta-schemas aside the official ones.

The meta-schema approach has a limitation. *An embedded schema can declare a different dialect*, so the premise is not fully maintained.

If the library gives **dialect conformance** only, the meta-schema approach seems like the right shape. Arguably the library doesn’t need to ship one at all, since users can already assemble it themselves.
If the goal is to have a different compiler behavior, then the meta-schema cannot reach that behavior for the problem I raised earlier.

Also, the strict validation is opt and disabled by default, so the additional cost is only incurred when this strict validation is explicitly enabled.

